Issue #5, opened 2021-01-05
Test all matchers from jest-dom


calebeby:
- [ ] `toBeDisabled`
- [ ] `toBeEnabled`
- [ ] `toBeEmpty`
- [ ] `toBeEmptyDOMElement`
- [ ] `toBeInTheDocument`
- [ ] `toBeInvalid`
- [ ] `toBeRequired`
- [ ] `toBeValid`
- [ ] `toBeVisible`
- [ ] `toContainElement`
- [ ] `toContainHTML`
- [ ] `toHaveAttribute`
- [ ] `toHaveClass`
- [ ] `toHaveFocus`
- [ ] `toHaveFormValues`
- [ ] `toHaveStyle`
- [ ] `toHaveTextContent`
- [ ] `toHaveValue`
- [ ] `toHaveDisplayValue`
- [ ] `toBeChecked`
- [ ] `toBePartiallyChecked`
- [ ] `toHaveDescription`

---

- I think these should be in the `tests` folder rather than the `examples` folder. I think we should try to only have more realistic, complete, useful tests in `examples`.
- For matchers that fail, we should use `test.skip(` when we make PR's so that we can avoid making the full test suite fail.

PR #10, opened 2021-01-08
More matcher tests

emersonthis:
This is the first half of the matcher tests (#5). These tests are more substantive, so I thought it best to merge in parts to avoid a monster PR.


calebeby:
@emersonthis What do you think about testing `.not` as well?

tests/jest-dom-matchers/toBeInTheDocument.test.ts
```diff
@@ -1,10 +1,10 @@
...
-	await expect(inDoc).toBeInTheDocument();
-	await inDoc.parentNode.removeChild(inDoc);
-	await expect(expect(inDoc).toBeInTheDocument()).rejects.toThrow('foo');
-});
\ No newline at end of file
+  const { screen, utils } = await createTab();
+  await utils.injectHTML(`<button>Hi</button>`);
+  const inDoc = await screen.getByText('Hi');
+  await expect(inDoc).toBeInTheDocument();
+  await inDoc.parentNode.removeChild(inDoc);
```
calebeby, line 8: This won't work since we don't have synchronous DOM access. You'll have to use [`.evaluate`](https://playwright.dev/docs/api/class-jshandle#jshandleevaluatepagefunction-arg):

```suggestion
  await inDoc.evaluate((el) => el.remove())
```

tests/jest-dom-matchers/toBeValid.test.ts
```diff
@@ -0,0 +1,38 @@
+import { createTab } from 'test-mule';
+
+test('toBeValid', async () => {
```
calebeby, line 3: @emersonthis What do you think about combining this test with `toBeValid` since it is very similar?

tests/jest-dom-matchers/toBeInTheDocument.test.ts
```diff
@@ -0,0 +1,10 @@
+import { createTab } from 'test-mule';
+
+test.skip('toBeInTheDocument', async () => {
```
calebeby, line 3: I fixed the bug on `main` that caused this matcher to not work, it should work now

PR #16, opened 2021-02-02
Add tests for the rest of the jest-dom matchers and fix a bunch of bugs

calebeby:


src/jest-dom/to-have-style.js
```diff
@@ -0,0 +1,79 @@
...
+
+function printoutStyles(styles) {
+  return Object.keys(styles)
+    .sort()
+    .map((prop) => `${prop}: ${styles[prop]};`)
+    .join('\n');
+}
+
+/** @this {jest.MatcherUtils} */
+export function toHaveStyle(htmlElement, css) {
```
calebeby, line 37: I copy-pasted this implementation and made a bunch of changes because the original implementation had dependencies which didn't want to bundle

emersonthis (approved):

PR #17, opened 2021-02-02
Add tests for utils

calebeby:
I added tests for almost all the utils (loadCSS, injectHTML, runJS, etc.)

Along the way, I noticed that the error messages from runJS were really unhelpful (because the messages originated from the browser). So I implemented a thing that uses source maps to find the original location of the error-producing code. I am pleased with how this turned out:

image

And here is an example with transpiled JSX:

image

It also handles source-mapping errors coming from external files, if you import an external file from within the runJS code string.

So glad that source maps exist 🙌 

For context, here is the error stack that appears on `main`:

image


package.json
```diff
@@ -23,16 +23,19 @@
...
     "rollup-plugin-terser": "^7.0.2",
+    "sass": "^1.32.6",
     "smoldash": "^0.9.0",
+    "source-map": "^0.7.3",
     "typescript": "^4.1.3"
   },
   "dependencies": {
     "@types/testing-library__jest-dom": "^5.9.5",
-    "puppeteer": "^5.5.0",
+    "puppeteer": "^6.0.0",
```
calebeby, line 38: They released this today and it uses the first Chromium version that supports top-level await, which is convenient because that's just what I needed [here](https://github.com/cloudfour/test-mule/pull/17/files#diff-91fbb1d3b9ae2a143e68e5008ce5697d5fb6a4979da129fd57c8c219d90d66ceR55)

PR #22, opened 2021-02-08
Add menu example

calebeby:


emersonthis:
Same question. I think we covered this in our meeting this morning, but let me know if you still want comments on specific code.

Issue #19, opened 2021-02-05
Rename debug()


calebeby:
Emerson suggested renaming it. I'm wondering if it instead makes sense to change the functionality to be triggered by returning false from the test function:

```js
test(
  "something",
  withBrowser.headed(async ({ screen, utils }) => {
    // ...
    return false;
	// ... lines to not run
  })
);
```

Benefits of this approach:
- No need to add/remove `debug` from the destructured parameter list
- Still allows you to stop/debug the test in the middle because you can put the return statement before other code that you don't want to run

Issue #15, opened 2021-02-01
Worker does not "gracefully exit" when one test suite fails when there are multiple suites


calebeby:
> A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.

This happens if I run >1 test file and >=1 fails. It does not happen if I only run one test file.

PR #23, opened 2021-02-10
Change from debug() to `return false`, and fix browser cleanup bug

calebeby:
Fixes https://github.com/cloudfour/test-mule/issues/15
Fixes https://github.com/cloudfour/test-mule/issues/19

For #19:

Reproduction (error about unclosed instances should show up on `main`, most of the time, and never on this branch)

```js
import { withBrowser } from 'test-mule';

test(
  'passing',
  withBrowser(async ({ page }) => {
    await page.goto('https://google.com')
  }),
);

test(
  'failing',
  withBrowser(async ({ page }) => {
    expect(4).toEqual(43424);
  }),
);
```

For #15:

```js
import { withBrowser } from 'test-mule';

test(
  'passing',
  withBrowser(async ({ page }) => {
    await page.goto('https://google.com')
    return false
  }),
);
```

`return false` tells `test-mule` to leave the browser open, even if the test passes (but only for headed browsers; headless browsers are always closed)

Diff without whitespace: https://github.com/cloudfour/test-mule/pull/23/files?diff=split&w=1

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -45,15 +45,14 @@ export interface TestContext {
   /** Returns DOM Testing Library queries that only search within a single element */
   within(element: puppeteer.ElementHandle | null): BoundQueries;
   page: puppeteer.Page;
-  /** Stops the test and leaves the browser open for debugging */
-  debug(): void;
 }
 
 let serverPromise: Promise<vite.ViteDevServer>;
 
 interface WithBrowserBase {
   (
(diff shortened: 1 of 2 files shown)


emersonthis:
Did our verbal conversation from earlier today give you what you need?

calebeby:
Oops I forgot we were going to change this to have an option in the configure object. I'll make another PR

PR #29, opened 2021-02-17
Add README

calebeby:
Rendered preview: https://github.com/cloudfour/test-mule/blob/readme/README.md

Still TODO in README:

- [x] Device Emulation
- [x] Debugging/troubleshooting
- [x] Switching to headed browser
- [x] Using `ElementHandle.evaluate(...)`
- [ ] ~Making Good Abstractions in your tests - section mostly for Cypress users wanting an equivalent for custom commands, and for people wanting to use beforeEach/afterEach~ This should be a blog post
- [x] Table of Contents

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,13 @@
 # Test Mule
 
-Test Mule is a library that allows you to use real browsers in your Jest tests. Test Mule is focused on helping you write tests that are [as similar as possible to how users use your application](https://twitter.com/kentcdodds/status/977018512689455106). It is built on [Puppeteer](https://github.com/puppeteer/puppeteer), [Testing Library](https://testing-library.com), and [jest-dom](https://github.com/testing-library/jest-dom).
+Test Mule is a library that allows you test web applications using real browsers in your Jest tests. Test Mule is focused on helping you write tests that are [as similar as possible to how users use your application](https://twitter.com/kentcdodds/status/977018512689455106).
 
-This exists because there is not an existing solution for browser testing that uses a real browser, supports [Testing Library](https://testing-library.com) out of the box, and integrates with Jest and non-browser tests.
+Test Mule is driven by these goals:
+
+- Use a real browser so that the test environment is much closer to what a real user uses.
+- Integrate with existing tests - browser testing should not be a "separate thing" from other tests.
+- Build on top of existing [Testing Library](https://testing-library.com) tools.
+- Make the testing experience as fast and painless as possible.
 
 - [Usage](#usage)
   - [Getting Started](#getting-started)
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Test Mule is a library that allows you to use real browsers in your Jest tests. Test Mule is focused on helping you write tests that are [as similar as possible to how users use your application](https://twitter.com/kentcdodds/status/977018512689455106). It is built on [Puppeteer](https://github.com/puppeteer/puppeteer), [Testing Library](https://testing-library.com), and [jest-dom](https://github.com/testing-library/jest-dom).
 
-This exists because there is not an existing solution for browser testing that uses a real browser, supports Testing Library out of the box, and integrates with Jest and non-browser tests.
+This exists because there is not an existing solution for browser testing that uses a real browser, supports [Testing Library](https://testing-library.com) out of the box, and integrates with Jest and non-browser tests.
 
 ## Usage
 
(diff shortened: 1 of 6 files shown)


calebeby:
@emersonthis feel free to take a look / review

README.md
```diff
@@ -1,44 +1,428 @@
...
+  'test name',
+  withBrowser(async () => {
+    // Your test code here
+  }),
+);
+```
+
+### Loading Content
+
+#### Option 1: Navigating to a real page
```
emersonthis, line 26: Do we recommend this? If this isn't the preferred / expected way for people to use this tool, I recommend moving it off the front lines of our documentation and instead maybe just mention it's a supported alternative (and link to further documentation elsewhere).

README.md
```diff
@@ -1,44 +1,428 @@
...
+```js
+import { withBrowser } from 'test-mule';
+
+test(
+  'loading css with utils.loadCSS',
+  withBrowser(async ({ utils }) => {
+    // ... Whatever code you had before to load your content
+
+    // You can import CSS (or Sass, Less, etc.) files
+    await utils.loadCSS('./something.sass');
```
emersonthis, line 116: scss is way more common these says (instead of sass). I would switch it to show `.css` in the code and mention support for the pre-processed languages in the comment. I think that's a little more intuitive than the inverse.

README.md
```diff
@@ -1,44 +1,428 @@
...
+The Testing Library queries are exposed through the [`screen` property](https://github.com/cloudfour/test-mule/blob/readme/README.md#testmulecontextscreen) in the test context parameter.
+
+**You must `await` the result of the query. This is necessary to accomodate for the asynchronous communication with the browser.**
+
+```js
+import { withBrowser } from 'test-mule';
+
+test(
+  'selecting elements example',
+  withBrowser(async ({ utils, screen }) => {
```
emersonthis, line 134: Maybe just add:
```
await utils.injectHTML('<button>Log In</button>');
```

README.md
```diff
@@ -1,44 +1,428 @@
...
+    const button = await screen.getByText();
+    // jest-dom matcher -- Runs in browser, *must* be awaited
+    await expect(button).toBeVisible();
+    // Built-in Jest matcher -- Runs only in Node, does not need to be awaited
+    expect(5).toEqual(5);
+  }),
+);
+```
+
+## Comparisons with other testing tools/Why does this exist?
```
emersonthis, line 414: This is great. And this location feels about right for this detailed look at the alternatives. I feel like we should also split off a 1-3 sentence version of answering the "Why?" way up at the top.

README.md
```diff
@@ -1,44 +1,428 @@
-# test-mule
+# Test Mule
 
-[not ready for use]
+Test Mule is a library that allows you to use real browsers in your Jest tests. Test Mule is focused on helping you write tests that are [as similar as possible to how users use your application](https://twitter.com/kentcdodds/status/977018512689455106). It is built on [Puppeteer](https://github.com/puppeteer/puppeteer), [Testing Library](https://testing-library.com), and [jest-dom](https://github.com/testing-library/jest-dom).
```
emersonthis, line 3: There's a comment below about discussing the "Why?" earlier. After returning to the beginning of this document, it could could probably be worked into this paragraph somehow. Right now this accurately describes what Test Mule is but doesn't really spell out _why_ it is or answer the reader's likely first question: _Why should I care about this? (...among the other testing tools I already know)_.  I think the easiest way to make that connection is to speak to the audience's "pain". For inspiration, [this new js framework](https://markojs.com/) did a pretty good job of catching my attention, even though I'm **not** in the market for a new js framework. I think it worked on me because they quickly touched on pain points that I recognize.

README.md
calebeby: It is how people use Cypress. I think it is a valid use case.

README.md
calebeby: I pushed a change, let me know if there are any suggestions for stuff to add or change for the Why

README.md
emersonthis: I'm not questioning it's validity at all. I'm just wondering if it's what we want to lead with. Listing it first in this way gives me the impression that it's the default / most-canonical method. Maybe you feel that's true? 

Part of why I think this is worth the bother if that this use case seems least representative of why TM is cool and special (IMO). If the reader stopped here, they'd be forgiven for thinking we'd sort of rebuilt Cypress, right? To me, the improved ergonomics are a **big** part of what's special about TM... do you agree? <-- this is the part I think is most interesting about this conversation.

README.md
emersonthis: I spotted some changes lower down. But at the moment, nothing on line 3 speaks to _why_.  As an exercise, I recommend trying to answer this question directly in 3 sentences or less: "As a developer, why should I care about Test Mule?" (and then incorporate that above)

Another exercise that might help: "As someone who doesn't know much about testing, why should I care about Test Mule?" 

Let's chat tomorrow if what I'm nudging you to do isn't making sense.

README.md
calebeby: I reordered the 3 options. The "navigate to real page" option is now third.

README.md
calebeby: @emersonthis I rewrote the first couple paragraphs, mind taking another look? I think I've hit the main points of why Test Mule exists and why people should care about it, both for people who are new to browser testing, and for people who have experience with other tools.

emersonthis: I like it!

emersonthis (approved):

PR #43, opened 2021-02-26
Handle forgotten await

### LINKED ISSUE #30: Try to catch errors where people forget to await
It is probably possible to catch errors that happen when people forget to await for some (or maybe all) of these cases:

- Execute jest-dom matcher without `await`
- Execute utilities (runJS, etc.) without `await`
- Run user methods without `await`
- Execute Testing Library queries without `await`
- Run `page.*` methods without `await` (this one will be harder, may not be possible)

We can probably catch those errors and provide a useful error message so people are not confused

calebeby:
Provide a helpful message if the user forgets to use `await`,

For example, if a user forgets to use `await` in the jest-dom assertion:

```js
test(
  'example',
  withBrowser(async ({ screen, utils }) => {
    await utils.injectHTML('<button>Hi</button>');
    const button = await screen.getByText(/hi/i);
    expect(button).toBeVisible();
  }),
);
```

Then a useful error message is produced:

```
Cannot execute assertion toBeVisible after test finishes. Did you forget to await?

  103 |     await utils.injectHTML('<button>Hi</button>');
  104 |     const button = await screen.getByText(/hi/i);
> 105 |     expect(button).toBeVisible();
      |                    ^
  106 |   }),
  107 | );
  108 |
```

This is also handled for utility functions, user methods, and Testing Library queries.

Closes #30

### DIFF
--- a/.changeset/heavy-emus-walk.md
+++ b/.changeset/heavy-emus-walk.md
@@ -2,7 +2,7 @@
 'test-mule': minor
 ---
 
-Provide a helpful message if the user forgets to use `await`,
+Provide a helpful message if the user forgets to use `await`.
 
 For example, if a user forgets to use `await` in the jest-dom assertion:
 
--- a/.changeset/heavy-emus-walk.md
+++ b/.changeset/heavy-emus-walk.md
@@ -1,0 +1,34 @@
+---
+'test-mule': minor
+---
+
+Provide a helpful message if the user forgets to use `await`,
+
+For example, if a user forgets to use `await` in the jest-dom assertion:
+
+```js
+test(
+  'example',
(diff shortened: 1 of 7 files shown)


.changeset/heavy-emus-walk.md
```diff
@@ -0,0 +1,34 @@
+---
+'test-mule': minor
+---
+
+Provide a helpful message if the user forgets to use `await`,
```
gerardo-rodriguez, line 5: Is this supposed to be a period? 🤔 
```suggestion
Provide a helpful message if the user forgets to use `await`.
```

gerardo-rodriguez (approved):
LGTM 👍

PR #45, opened 2021-03-02
Add ability to pass arguments to browser in runJS

### LINKED ISSUE #28: Allow passing variables into runJS
Right now there is no way to pass variables from node into the browser using runJS. We should have a mechanism for that.

---

From the original issue:

This may be a bad idea but I wanted to write it down before I forget it.

```js
import { withBrowser } from "test-mule";

test(
  "example",
  withBrowser(async ({ utils, screen }) => {
    const button = await screen.getByText(/button/i);
    await utils.runJS`
      console.log(${button})
    `;
  })
);
```

This would work by receiving each interpolation and passing it into page.evaluate. We'd have to implement something so that the source-mapped-errors thing still finds the code string in the source code, but should be doable since we know the static chunks

calebeby:
Adds ability to pass variables to the browser in runJS:

```js
import { withBrowser } from 'test-mule';
test(
  'runJS example with argument',
  withBrowser(async ({ utils, screen }) => {
    // element is an ElementHandle (pointer to an element in the browser)
    const element = await screen.getByText(/button/i);
    // we can pass element into runJS and the default exported function can access it as an Element
    await utils.runJS(
      `
        export default (element) => console.log(element);
      `,
      [element],
    );
  }),
);
```

Closes #28 

In #28 I mentioned other syntax options for how to make this possible, open to feedback 👀

### DIFF
--- a/.changeset/proud-numbers-grin.md
+++ b/.changeset/proud-numbers-grin.md
@@ -1,0 +1,24 @@
+---
+'test-mule': minor
+---
+
+Add ability to pass variables to the browser in runJS:
+
+```js
+import { withBrowser } from 'test-mule';
+
+test(
+  'runJS example with argument',
+  withBrowser(async ({ utils, screen }) => {
+    // element is an ElementHandle (pointer to an element in the browser)
+    const element = await screen.getByText(/button/i);
+    // we can pass element into runJS and the default exported function can access it as an Element
+    await utils.runJS(
+      `
+        export default (element) => console.log(element);
+      `,
+      [element],
+    );
+  }),
+);
+```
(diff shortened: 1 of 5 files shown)


emersonthis:
@calebeby Should the test make an assertion to ensure that `element` is passed as we expect to the browser? As written, I think this test will fail with a ReferenceError if `element` isn't defined at all. But I think it would pass if it's anything else right? Even `null` or `undefined`? Is that worth checking for?

calebeby:
@emersonthis that was just an example I have from the changelog. I forgot to write an actual test, so I added one here: [`cd8d4d8`](https://github.com/cloudfour/test-mule/pull/45/commits/cd8d4d8fa595b9657c34eeda9504502463c7e46b)

emersonthis (approved):

PR #58, opened 2021-05-05
Implement `user.clear()`

calebeby:
This PR adds the `user.clear` method from [`user-event`](https://github.com/testing-library/user-event#clearelement)

### DIFF
--- a/.changeset/afraid-pigs-cry.md
+++ b/.changeset/afraid-pigs-cry.md
@@ -1,0 +1,7 @@
+---
+'test-mule': minor
+---
+
+Implement `user.clear()`
+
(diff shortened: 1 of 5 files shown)


.changeset/afraid-pigs-cry.md
```diff
@@ -0,0 +1,7 @@
+---
```
Paul-Hebert, line 1: This is quite a file name!

Paul-Hebert (approved):
LGTM!

.changeset/afraid-pigs-cry.md
calebeby: image

PR #60, opened 2021-05-13
Implement `user.selectOptions()`

calebeby:
Reimplements https://github.com/testing-library/user-event#selectoptionselement-values using Puppeteer methods, and adds it to the `user` object


spaceninja (approved):
I'm not familiar enough with this code to be 100% confident in approving it, but the changes look reasonable.

gerardo-rodriguez (approved):
Great work, @calebeby! 🎉

PR #61, opened 2021-05-13
Update puppeteer to v9.1.1

calebeby:


src/jest-dom/to-have-style.js
```diff
@@ -67,9 +67,11 @@ export function toHaveStyle(htmlElement, css) {
         );
       }
 
-      const receivedSubset = Object.fromEntries(
```
calebeby, line 70: This was a really annoying bug to figure out... Chrome changed how properties on CSSStyleDeclaration appear, now they are a part of the prototype so they don't appear in `Object.fromEntries` 😦 

https://chromestatus.com/feature/6464391237074944

spaceninja (approved):

PR #63, opened 2021-05-24
Update dependencies

calebeby:


jest.config.js
```diff
@@ -6,6 +6,10 @@ module.exports = {
   testRunner: 'jest-circus/runner',
   watchPathIgnorePatterns: ['<rootDir>/src/'],
   transform: {
-    '^.+\\.tsx?$': ['esbuild-jest', { sourcemap: true }],
+    '^.+\\.[jt]sx?$': ['esbuild-jest', { sourcemap: true }],
   },
+  // Don't transform node_modules, _except_ ansi-regex
+  // ansi-regex is ESM and since we are jusing Jest in CJS mode,
```
gerardo-rodriguez, line 12: Typo?
```suggestion
  // ansi-regex is ESM and since we are using Jest in CJS mode,
```

.changeset/many-carrots-join.md
```diff
@@ -0,0 +1,5 @@
+---
+'test-mule': patch
+---
+
+Update dependencies
```
gerardo-rodriguez, line 5: You'll have to explain these files to me at some point as I'm not clear what they do or why they have the file names they do. 😝 🙂

gerardo-rodriguez (approved):
Left one typo suggestion, otherwise, LGTM. 👍

.changeset/many-carrots-join.md
calebeby: They are changeset files, the file name is auto-generated! When you make a change that is user-affecting, you can run `npx changeset` and it will ask you questions and generate the file. The file is used to create the changelog.

Whenever there is at least one changeset file on `main`, there is a github action that creates a PR like [this](https://github.com/cloudfour/test-mule/pull/62) that aggregates all the changesets into the changelog and figures out the smallest version bump that will work for all the changesets. When you merge that PR, it will create a release and publish to npm.

I like this workflow a lot! Previously I used [semantic-release](https://semantic-release.gitbook.io/semantic-release/) but I don't like that so much because the changelog is generated from the commit messages

PR #78, opened 2021-05-28
Handle errors when browser launching fails

calebeby:
This PR fixes a bug I noticed on cloudfour-patterns.

Previously, if npm failed to run the postinstall script for puppeteer (or ignored it? I noticed npm@7 ignores postinstall scripts sometimes), then when you run pleasantest, here is what would happen:
```
Pleasantest sees that there is no cached browser instance to connect to
Pleasantest spawns a new process to launch a new browser
In the spawned process, puppeteer.launch() is called
puppeteer.launch() throws an error, saying that the browser binary is missing, but since it is in a child process, the error is not displayed anywhere
---
5 seconds later: Jest force-quits the test because it timed out
```

Now, errors that are thrown in the child process get passed to the main process, where they are thrown:

```
Failed to start browser: Could not find expected browser (chrome) locally. Run `npm install` to download the correct Chromium revision (869685).
```

## Testing

- `npm ci`
- Lose the browser binary: `rm -rf node_modules/puppeteer/.local-chromium`
- Clear out any references to cached browser instances: `rm /Users/<your username>/Library/Application\ Support/pleasantest-nodejs/config.json`
- Run a test (probably just one, running them all takes a long time if they are all failing): `npm run test toBeVisible`. It should fail and display a helpful message.
- Run the test again (or another test). It should fail again with the same message. (This is to make sure that it does not try to cache the browser that failed to launch the first time)
- `node ./node_modules/puppeteer/install.js` to re-download the browser binaries
- `npm run test` should pass

### DIFF
--- a/.changeset/wicked-pumpkins-join.md
+++ b/.changeset/wicked-pumpkins-join.md
@@ -1,0 +1,5 @@
+---
+'pleasantest': patch
+---
+
+Handle errors thrown by browser launcher instead of silently hanging
--- a/src/connect-to-browser.ts
+++ b/src/connect-to-browser.ts
@@ -121,12 +121,24 @@ export const connectToBrowser = async (
     detached: true,
     stdio: 'ignore',
   });
-  const wsEndpoint = await new Promise<string>((resolve) => {
+  const wsEndpoint = await new Promise<string>((resolve, reject) => {
     subprocess.send({ browser, headless });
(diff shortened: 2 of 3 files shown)


calebeby:
image

Easiest to review with "hide whitespace changes" checked

Paul-Hebert (approved):
Code looks good and works as advertised :+1:

PR #106, opened 2021-06-14
Cache browsers locally instead of globally

### LINKED ISSUE #77: Cached browser config should be stored in node_modules
Right now the websocket URL of the cached browser is stored globally which causes problems if there are multiple pleasantest projects with different puppeteer versions (and different browser versions). Also, it is frustrating when clearing node_modules does not clear all state.

By storing the configuration file with the cached browser in node_modules, we can solve both of these problems

calebeby:
Closes #77 

Now the browser instance cache file is stored in the pleasantest folder rather than globally (previously it was `/Users/<user>/Library/Application\ Support/pleasantest-nodejs/config.json`. Now it is stored as a sibling to the `dist` folder.

This solves several problems:

- If someone is having issues with cached browsers and wants to start fresh browsers instead of continuing to use whatever was wrong, now if they run `npm ci`, they will get fresh browsers
- If someone has multiple versions of puppeteer (/ multiple versions of pleasantest) installed, there are no longer issues with the cache returning a browser whose version is mismatched with puppeteer's version.


src/connect-to-browser.ts
```diff
@@ -87,28 +87,35 @@ export const connectToBrowser = async (
 ) => {
   // I acknowledge that this code is gross and should be refactored
   // Constraints:
-  // - If there is no browser in the config, multiple concurrent processes should only start 1 new browser
-  // - If there is a killed browser in the config, multiple concurrent processes should only start 1 new browser
-  // - If there "starting" in the config but nothing is really starting, multiple concurrent processes should only start 1 new browser
+  // - If there is no browser in the cache, multiple concurrent processes should only start 1 new browser
+  // - If there is a killed browser in the cache, multiple concurrent processes should only start 1 new browser
+  // - If there "starting" in the cache but nothing is really starting, multiple concurrent processes should only start 1 new browser
   // TODO: Idea: use a state machine!!!
```
calebeby, line 93: This code is still pretty gross, but it has held together way better than I expected. I'm leaving this comment here because I would still like to come back and rewrite it, but for the time being it works and is pretty resilient and "robust enough".

src/connect-to-browser.ts
```diff
@@ -1,32 +1,32 @@
...
 import * as path from 'path';
 import { promises as fs } from 'fs';
-import envPaths from 'env-paths';
 import * as puppeteer from 'puppeteer';
 // @ts-expect-error the bundle: syntax is from a plugin in the rollup config and TS does not know about it
 import startDisownedBrowserPath from 'bundle:./start-disowned-browser';
+import { fileURLToPath } from 'url';
 
-const readConfig = async (configPath: string) => {
+const readCache = async (cachePath: string) => {
```
calebeby, line 9: Renamed all the things config -> cache because cache is more descriptive of what it actually is. I think at one point I had planned to put other stuff than browser cache URL's in it, but not anymore.

gerardo-rodriguez (approved):
LGTM 👍

PR #115, opened 2021-06-22
Replace vite module server with wmr/custom module server

calebeby:
## Reviewing

It might be easiest to review in:
https://github1s.com/cloudfour/pleasantest/pull/115

To review the changes to `src/module-server/rollup-plugin-container.ts`, use [this diff](https://github.com/cloudfour/pleasantest/pull/115/commits/fd0d8ab7c0f52886d7aae0952e62049b89ef7a73#diff-2f4a092fc1dc99b13b08299964e49b8332085f7e085b5be5a74833faac119daf) which shows the differences from WMR's version of that file.

### Source Maps

Since WMR does not support source maps, Pleasantest's code to fix the stack traces from in-browser errors does not work. So I commented out the source-map-dependent code in Pleasantest for now. I will follow up with a PR to add source map support.

For now, errors from runJS/runJS-imported code look like this 🤮:

Screen Shot 2021-06-22 at 12 39 56 PM

I added `.skip` for the tests for this functionality, so in the follow-up PR I'll re-enable those tests.

### Configurability of module server

Right now the configuration is not exposed. This is a pretty easy feature to add but I figured I'd leave it for a follow-up PR since this PR is already really big, only to get the tests passing.

Proposed API:

```js
import { withBrowser, configure } from 'pleasantest'

// Global default configuration
configure({
  plugins: [
    /* some rollup plugins here */
  ],
  esbuild: {
    /* some option here */
  },
  // ...
})

test(
  'asdf',
  withBrowser({
    // overrides to global configuration (along with other options to withBrowser like device emulation)
  }, async () => {
    // test here
  }),
)
```


calebeby:
Note to self: TODO before merge: Add back tests for react/react-dom

src/module-server/middleware/js.ts
```diff
@@ -0,0 +1,109 @@
...
+  return async (req, res, next) => {
+    try {
+      // Normalized path starting with slash
+      const path = posix.normalize(req.path);
+      let id: string;
+      let file: string;
+      if (path.startsWith('/@npm/')) {
+        id = path.slice(1);
+        // TODO:
+        file = 'halp im stuck in a string';
```
calebeby, line 32: fix before merge

src/index.ts
```diff
@@ -298,47 +300,47 @@ const createTab = async ({
...
+    //     column = sourceLocation.column + 1;
+    //     line = sourceLocation.line + linesTillInlineCode;
+    //   } else {
+    //     column = sourceLocation.column + 1;
+    //     line = sourceLocation.line;
+    //   }
+
+    //   fileName = localFileName;
+    //   return `    at ${fileName}:${line}:${column}`;
+    // });
```
gerardo-rodriguez, line 343: Can you add a TODO comment to help explain why this code is commented out? 🙂

src/index.ts
```diff
@@ -350,10 +352,11 @@ const createTab = async ({
...
     const ErrorConstructor = specializedErrors[errorName] || Error;
     const error = new ErrorConstructor(message);
+    error.stack = stack;
 
-    error.stack = `${errorName}: ${message}\n${(
-      await Promise.all(modifiedStack)
-    ).join('\n')}`;
+    // Error.stack = `${errorName}: ${message}\n${(
+    //   await Promise.all(modifiedStack)
+    // ).join('\n')}`;
```
gerardo-rodriguez, line 359: Same here, can you add a TODO comment explaining why this is commented out?

src/module-server/plugins/npm-plugin.ts
```diff
@@ -0,0 +1,229 @@
...
+    fs.writeFile(cachePath, code),
+  );
+};
+
+const getFromCache = async (cachePath: string) => {
+  return (
+    npmTranspileCache.get(cachePath) ||
+    fs
+      .readFile(cachePath, 'utf8')
+      .catch(() => {})
```
gerardo-rodriguez, line 31: Is this empty function in here by design? If so, can you leave a comment explaining why? 🤔

tests/utils/runJS.test.tsx
```diff
@@ -107,11 +106,10 @@ test(
   }),
 );
 
-test(
+test.skip(
```
gerardo-rodriguez, line 109: Double-checking you meant to leave this in. 😉

tests/utils/runJS.test.tsx
```diff
@@ -143,10 +142,16 @@ test(
...
+    const error3 = await utils
+      .runJS(`console.log('hello)`)
+      .catch((error) => error);
+
+    expect(await printErrorFrames(error3)).toMatchInlineSnapshot();
   }),
 );
 
-test(
+test.skip(
```
gerardo-rodriguez, line 154: Same here, did you mean to leave this `skip` in? 🙂

gerardo-rodriguez (approved):
I left a few questions, otherwise, looks good to me. Quite the PR! 😅

tests/utils/runJS.test.tsx
calebeby: Yeah this fails because the source map code is commented out because the WMR based module loader doesn't support source maps. I'm opening a PR next week probably with adding the source map support back

PR #116, opened 2021-06-25
Add direct CSS loading and static file loading to module server

calebeby:
Before this PR, CSS was supported in this way:

You could use loadCSS:
```js
await utils.loadCSS('./something.sass') // or css, etc
```

You could use css imports from JS files:

```js
import './asdf.css'
```

The CSS file would be piped through PostCSS and converted into a JS script that injects the CSS as a <style> tag.

Also, sass/less/stylus/less were supported (via [rollup-plugin-postcss](https://github.com/egoist/rollup-plugin-postcss)):

```js
import './asdf.sass'
```

However, if you requested a sass/css file via HTTP, you would get back a JS file with the code to inject the CSS. This is not good.

Now, in addition to those use cases, CSS is supported at the HTTP server layer, so if you request a CSS file via something _other_ than a JS import, it will end up being a CSS file rather than a JS file.

This is done by adding `?import` to imports in JS files to distinguish between JS/CSS (same as Vite does, and similar to what WMR does):

```js
import './asdf.css' // => transpiles to import './asdf.css?import'
```

Then the HTTP server knows to return a JS file with the CSS injector if `?import` exists, and CSS directly otherwise.

An example use-case for this is if `@import` is used

Also, the HTTP server now supports static assets (like fonts/images loaded via CSS - which I ran into on cloudfour.com)


tests/utils/loadCSS.test.ts
```diff
@@ -50,3 +50,29 @@ test(
...
+  withBrowser(async ({ utils }) => {
+    // Note: this test doesn't actually test anything since we can't detect the background image loading via JS.
+    // This has to be manually tested with a headed browser
+    // Even without the headed browser, if this fails there will be a console.error in the node console (but it doesn't fail the test unfortunately)
+    await utils.injectHTML(`
+      <div>I have a background image</div>
+    `);
+    await utils.loadCSS('./external-with-reference.css');
+  }),
+);
```
gerardo-rodriguez, line 78: @calebeby Does `element.style.backgroundImage` do anthing to help this test? 🤔

calebeby: It won't tell us whether the image actually loaded

calebeby: Looking at this again, I think it could help gain confidence by making sure that the resolved URL is correct. I'll add it 👍

calebeby: I just remembered that we could probably use request interception for this 👀

gerardo-rodriguez (approved):
Nice! 🎉

PR #139, opened 2021-07-07
Improve error message when Promise is passed into jest-dom matcher

calebeby:
If you do this:

```js
await expect(screen.getByText(/asdf/))
```

That code is wrong because the `getByText` query is not `await`ed.

Previously, the error message in that case was along the lines of `elementHandle.asElement is not a function`. Now the error message is:

Screen Shot 2021-07-07 at 8 11 16 AM

### DIFF
--- a/src/extend-expect.ts
+++ b/src/extend-expect.ts
@@ -1,7 +1,12 @@
 import type { ElementHandle, JSHandle } from 'puppeteer';
 import { createClientRuntimeServer } from './module-server/client-runtime-server';
 import { deserialize, serialize } from './serialize';
-import { isPromise, jsHandleToArray, removeFuncFromStackTrace } from './utils';
+import {
+  isElementHandle,
+  isPromise,
+  jsHandleToArray,
+  removeFuncFromStackTrace,
+} from './utils';
 
 const methods = [
   'toBeInTheDOM',
@@ -45,11 +50,7 @@ expect.extend(
         ...matcherArgs: unknown[]
       ) {
         const serverPromise = createClientRuntimeServer();
-        if (
-          typeof elementHandle !== 'object' ||
-          // eslint-disable-next-line @cloudfour/typescript-eslint/no-unnecessary-condition
-          !elementHandle?.asElement?.()
-        ) {
+        if (!isElementHandle(elementHandle)) {
           // Special case: expect(null).not.toBeInTheDocument() should pass
           if (methodName === 'toBeInTheDocument' && this.isNot) {
             // This is actually passing but since it is isNot it has to return false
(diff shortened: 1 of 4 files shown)


src/extend-expect.ts
```diff
@@ -45,7 +45,11 @@ expect.extend(
         ...matcherArgs: unknown[]
       ) {
         const serverPromise = createClientRuntimeServer();
-        if (typeof elementHandle !== 'object' || !elementHandle?.asElement()) {
+        if (
+          typeof elementHandle !== 'object' ||
+          // eslint-disable-next-line @cloudfour/typescript-eslint/no-unnecessary-condition
+          !elementHandle?.asElement?.()
+        ) {
```
Paul-Hebert, line 52: Nit: I found this kind of difficult to parse this conditional when it was split across multiple lines with a comment in the middle. I wonder if something like this would be easier to read?

```suggestion
        // eslint-disable-next-line @cloudfour/typescript-eslint/no-unnecessary-condition
        const hasAsElementMethod = elementHandle?.asElement?.();
        if (typeof elementHandle !== 'object' || !hasAsElementMethod) {
```

I don't feel super strongly about this though, so if that doesn't work as expected or has other downsides feel free to ignore.

tests/extend-expect.test.ts
```diff
@@ -0,0 +1,42 @@
...
+test(
+  'throws useful error if Promise is passed',
+  withBrowser(async ({ screen, utils }) => {
+    await utils.injectHTML('<h1>Hi</h1>');
+    await expect(expect(screen.getByText('Hi')).toBeVisible()).rejects
+      .toThrowErrorMatchingInlineSnapshot(`
+                  "[2mexpect([22m[31mreceived[39m[2m).toBeVisible()[22m
+
+                  [31mreceived[39m value must be an HTMLElement or an SVGElement.
+                  Received a [31mPromise[39m. Did you forget to await?"
```
Paul-Hebert, line 15: Out of curiosity, what's this syntax with the `[31m` stuff?

Paul-Hebert (approved):
Code and tests look good! Left one non-blocking nit inline.

This is a great DX improvement! I've been bitten by forgetting `await` in Stencil  E2E tests and the error there is much less helpful 😅

tests/extend-expect.test.ts
calebeby: Those are the terminal escape codes for the colored text

Paul-Hebert: Oh, interesting! TIL, thanks!

src/extend-expect.ts
calebeby: I extracted an `isElementHandle` function

Paul-Hebert: Oh perfect, thanks

PR #141, opened 2021-07-07
Refactor and improve browser stack trace printing

calebeby:
There are three changes to the way from-browser stack traces are printed now:

1. **Paths are always printed as absolute paths now**. Jest will only print the code frame for absolute paths. Previously, we were printing absolute paths for the lines that Jest would make code frames for, and relative paths otherwise. But that is more complicated and not necessary, because Jest will rewrite the absolute paths to relative paths anyways.
2. **URL's from node_modules get rewritten**. Previously, stack lines similar to `http://localhost:2345/@npm/react` would be printed as-is (since we don't generate source maps for node_modules). But when that happened, Jest would not be able to parse the stack trace (because it doesn't understand URL's, only file paths), and that would prevent it from displaying a code frame for _any_ line in the error. Now the `@npm` URL's get rewritten to absolute node_modules paths.
3. **Function names always are displayed if they exist**: Previously I had mistakenly thought that when function names (like `at FuncName (filePath:0:0)`) were one of the things that Jest's error stack parser was not able to recognize. It turns out that it _is_ able to recognize those, so now the code for conditionally displaying the function names depending on code frames has been removed.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -349,14 +349,13 @@ const createTab = async ({
       const mappedColumn = sourceLocation.column + 1;
       const mappedLine = sourceLocation.line;
       const mappedPath = sourceLocation.source || url.pathname;
-      const stackLine = printStackLine(
+      return printStackLine(
         // For the code frames, Jest will only recognize absolute paths
         path.join(process.cwd(), mappedPath),
         mappedLine,
         mappedColumn,
         stackItem.name,
       );
-      return stackLine;
     });
     const errorName = stack.slice(0, stack.indexOf(':')) || 'Error';
     const specializedErrors = {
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -335,20 +349,14 @@ const createTab = async ({
...
-      isFirst = false;
-      return `    at ${location}`;
+      const stackLine = printStackLine(
+        // For the code frames, Jest will only recognize absolute paths
+        path.join(process.cwd(), mappedPath),
+        mappedLine,
+        mappedColumn,
+        stackItem.name,
+      );
+      return stackLine;
```
Paul-Hebert, line 359: Nit: do we need to define this and return it on separate lines?

GitHub's suggestion tool isn't working for me, but could we just do `return printStackLine(...)`?

Paul-Hebert (approved):
Code looks good! I'll admit I don't have enough context to fully understand the underlying changes, but all of the changed lines look good and make sense in isolation. Let me know if you'd like any manual testing done on any of this.

src/index.ts
calebeby: Yeah I used to have a line in between but it is gone now, good catch

PR #142, opened 2021-07-07
Fix bug where css files in node_modules were pre-bundled

calebeby:
When CSS files were imported, from node_modules (like `import '@reach/dialog/styles.css'`), the CSS file would get pre-bundled into a JS file, which was causing issues because when it got imported the 2nd time it would try to parse the JS file as a CSS file and convert it to JS a second time. This makes it so that for any non-JS files that are imported from node_modules, the pre-bundle step is skipped.

### DIFF
--- a/.changeset/wet-snails-unite.md
+++ b/.changeset/wet-snails-unite.md
@@ -1,0 +1,5 @@
+---
+'pleasantest': patch
+---
+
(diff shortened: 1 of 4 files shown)


gerardo-rodriguez (approved):
Code LGTM! 👍

PR #143, opened 2021-07-07
Allow resolving subfolders with package.json files, and improve CJS interop

calebeby:
There are two changes in here, but they overlapped in their code changes a little bit so it was easiest to open it as one PR instead of dealing with conflicts.

1. **Resolving folders with package.json**: If I do `import 'preact/hooks'`, that already works (before this PR), because Preact uses an exports map, which tells us how to resolve the `./hooks` part. But many packages don't use exports maps yet, so they instead use the "old version" of this: in this case, they would create a `hooks` folder with a nearly-empty `package.json` in it, with a `main` or `module` field pointing to the file that they want to import. Before this PR, pleasantest didn't support this (it only traversed through the main package.json, e.g. `node_modules/preact/package.json`). Now, if an import resolves to a folder, recursion happens to resolve through a package.json in that folder (if it exists).
2. **Improved CJS interop**: There are packages on npm that are CJS-only, and they make it really really hard to statically determine their exports. Because of our bundling step, we need to be able to statically determine named exports of packages at compile-time. For some packages (`react`, `react-dom`, `prop-types`), they are really difficult to determine. Here's what the entry points for react and react-dom look like:
   ```js
   'use strict';
   if (process.env.NODE_ENV === 'production') {
     module.exports = require('./cjs/react.production.min.js');
   } else {
     module.exports = require('./cjs/react.development.js');
   }
   ```
   Obviously, by just looking at that file, you can't figure out which names are exported. #115 implemented a rollup transform that would detect the `module.exports = require(...` lines, and would then read the require'd files and statically determine the exports from those. But for `prop-types`, it doesn't work because there is a function call which makes the static analysis of the require'd file break:
	  ```js
	  if (process.env.NODE_ENV !== 'production') {
	    var throwOnDirectAccess = true;
	    module.exports = require('./factoryWithTypeCheckers')(ReactIs.isElement, throwOnDirectAccess);
	  } else {
	    module.exports = require('./factoryWithThrowingShims')();
	  }
	  ```
   In this PR, I deleted the static analysis for `module.exports = require(...`, and replaced it with an include-list of modules that should be `require`'d (at compile time) to determine their exports. This is used for `prop-types`, `react`, and `react-dom`. We can easily add more to that list down the road. If facebook eventually decides to add ESM to their packages, this should continue working without the compile-time `require` because it checks if the package is CJS before requiring it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "pleasantest",
   "version": "0.6.1",
   "engines": {
-    "node": "12 || 14 || 16"
+    "node": "^12.2 || 14 || 16"
   },
   "files": [
     "dist"
@@ -38,6 +38,7 @@
     "polka": "0.5.2",
     "preact": "10.5.14",
     "prettier": "2.3.2",
+    "prop-types": "^15.7.2",
     "react": "17.0.2",
     "react-dom": "17.0.2",
     "remark-cli": "9.0.0",
--- a/src/module-server/plugins/npm-plugin.ts
+++ b/src/module-server/plugins/npm-plugin.ts
@@ -233,7 +233,7 @@ const bundleNpmModule = async (mod: string, id: string, optimize: boolean) => {
     treeshake: true,
     preserveEntrySignatures: 'allow-extension',
     plugins: [
-      namedExports.length > 0 &&
+      hasSyntheticNamedExports &&
         ({
           // This plugin handles special-case packages whose named exports cannot be found via static analysis
           // For these packages, the package is require()'d, and the named exports are determined that way.
--- a/src/module-server/plugins/npm-plugin.ts
+++ b/src/module-server/plugins/npm-plugin.ts
@@ -1,14 +1,14 @@
 import { dirname, join, normalize, posix } from 'path';
 import type { Plugin, RollupCache } from 'rollup';
 import { rollup } from 'rollup';
-import { existsSync, promises as fs } from 'fs';
+import { promises as fs } from 'fs';
 import { resolve, legacy as resolveLegacy } from 'resolve.exports';
 import commonjs from '@rollup/plugin-commonjs';
 import { processGlobalPlugin } from './process-global-plugin';
 import * as esbuild from 'esbuild';
 import { parse } from 'cjs-module-lexer';
-import MagicString from 'magic-string';
 import { fileURLToPath } from 'url';
+import { createRequire } from 'module';
 
 // This is the folder that Pleasantest is installed in (e.g. <something>/node_modules/pleasantest)
 const installFolder = dirname(dirname(dirname(fileURLToPath(import.meta.url))));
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "pleasantest",
   "version": "0.6.1",
   "engines": {
-    "node": "12 || 14 || 16"
+    "node": "^12.2 || 14 || 16"
```
calebeby, line 5: `createRequire` is only supported by node 12.2+

calebeby:
BTW: The CJS interop stuff has tests for it, but the resolution stuff does not.

In a later PR, I'm planning to separate two files out of the npm plugin: 1 for the bundler thing that calls rollup, and one for the nodeResolve function. Then I'll write isolated tests for the nodeResolve function (and probably fix more edge cases)

src/module-server/plugins/npm-plugin.ts
```diff
@@ -133,31 +132,63 @@ const nodeResolve = async (id: string, root: string) => {
...
-      if (existsSync(path)) return { path, idWithVersion };
+      const stats = await fs.stat(path).catch(() => null);
+      if (stats) {
+        if (stats.isFile()) return { path, idWithVersion };
+        if (stats.isDirectory()) {
+          // If you import some-package/foo and foo is a folder with a package.json in it,
+          // resolve main fields from the package.json
+          const result = await resolveFromFolder(path, '.', packageName);
+          if (result) return { path: result.path, idWithVersion };
+        }
```
Paul-Hebert, line 149: Not sure how concerned we are with this in this project but I figured I'd flag this lint warning in case you didn't see it ☝️

src/module-server/plugins/npm-plugin.ts
```diff
@@ -166,58 +197,60 @@ let npmCache: RollupCache | undefined;
...
-              }
-
-              out.overwrite(
-                match.index,
-                re.lastIndex,
-                leadingWhitespace + replacement + trailingWhitespace,
-              );
-            } catch {
-              return;
+      namedExports.length > 0 &&
```
Paul-Hebert, line 236: Nit: this might be easier to read if you moved this logic outside of this config object.

Pseudo-code:

```js
const plugins = [];

if (namedExports.length > 0) {
  plugins.push({/* ... */});
}

const bundle = await rollup({
  /* ... */
  plugins,
  /* ... */
})
```

This is subjective though. I know the `&&` pattern is used pretty extensively in JSX so may be more familiar to other devs besides me.

Paul-Hebert (approved):
The code all looks good to me, but I feel like I'm lacking some context into how this interacts with the rest of Pleasantest. Depending on how thorough of a review you need you may want to tag in another dev for a second pair of eyes.

gerardo-rodriguez (approved):
I also don't have a full grasp of everything but code LGTM! 👍

src/module-server/plugins/npm-plugin.ts
calebeby: I think I might disable that warning, but thanks for flagging it

src/module-server/plugins/npm-plugin.ts
calebeby: Yeah I think it's pretty common in config arrays like this (babel plugins, rollup plugins, etc.)

PR #146, opened 2021-07-08
Allow throwing things other than errors within withBrowser

calebeby:
This is a super minor thing but sometimes for debugging I use the trick [shown here](https://github.com/cloudfour/pleasantest#withbrowser) (`throw new Error('leave the browser open')`) to leave the headed browser open even if the test passes. This is fine, but sometimes I want to type less, so I do `throw 0` which should have the same effect (makes the test fail so the browser says open). But there is a bug where Pleasantest expects anything thrown from the test to be an Error. This PR fixes that bug.

### DIFF
--- a/.changeset/kind-trainers-sleep.md
+++ b/.changeset/kind-trainers-sleep.md
@@ -1,0 +1,5 @@
+---
+'pleasantest': minor
+---
+
+Allow throwing things other than errors within withBrowser
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,11 @@
 {
   "name": "pleasantest",
-  "version": "0.6.1",
+  "version": "0.6.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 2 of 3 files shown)


Paul-Hebert (approved):
Looks good!

A couple non-blocking thoughts:

1. Would it be helpful to update the README to suggest `throw 0` since that's so concise?
2. Is this a common enough pain point that it would make sense to add a more formal way to do this? Like a `leaveOpen()` method or something?

calebeby:
> Is this a common enough pain point that it would make sense to add a more formal way to do this? Like a leaveOpen() method or something?
There used to be a `debug()` method, but I didn't like it because it had to be added to the destructured parameter list (which VSCode won't add automatically like it does for imports), and then it had to be removed from the parameter list too. I think I like relying on something that can be written without any imports/destructurings.

I think at one point there was a modifier like `withBrowser.leaveOpen` or something that I experimented with. What do you think of that?

PR #159, opened 2021-07-13
Node resolver: Refactor, add tests, and improve edge cases

### LINKED ISSUE #155: Add unit tests for node resolver and fix edge case bugs

calebeby:
Closes #155 

I went through https://nodejs.org/api/modules.html#modules_all_together, and tried to copy Node's commonJS resolution pretty closely.

Now the resolver is shared between the "extension resolver" and the "node modules resolver". Previously they did not share code, but since there is a lot of overlap, now they live in the same place and share code.

The "extension resolver" handles resolving `./asdf` to `./asdf.ts`, `./asdf/index.js`, and reading `./asdf/package.json` and resolving through there if it exists.

The "node modules resolver" handles resolving node_modules paths.

### DIFF
--- a/src/module-server/bundle-npm-module.ts
+++ b/src/module-server/bundle-npm-module.ts
@@ -5,6 +5,7 @@ import commonjs from '@rollup/plugin-commonjs';
 import { processGlobalPlugin } from './plugins/process-global-plugin';
 import * as esbuild from 'esbuild';
 import { parse } from 'cjs-module-lexer';
+// @ts-expect-error @types/node@12 doesn't like this import
 import { createRequire } from 'module';
 import { isBareImport, npmPrefix } from './extensions-and-detection';
 let npmCache: RollupCache | undefined;
--- a/src/module-server/bundle-npm-module.ts
+++ b/src/module-server/bundle-npm-module.ts
@@ -1,0 +1,117 @@
+import type { Plugin, RollupCache } from 'rollup';
+import { rollup } from 'rollup';
+import { promises as fs } from 'fs';
+import commonjs from '@rollup/plugin-commonjs';
+import { processGlobalPlugin } from './plugins/process-global-plugin';
+import * as esbuild from 'esbuild';
+import { parse } from 'cjs-module-lexer';
+import { createRequire } from 'module';
+import { isBareImport, npmPrefix } from './extensions-and-detection';
+let npmCache: RollupCache | undefined;
+
+/**
+ * Any package names in this set will need to have their named exports detected manually via require()
+ * because the export names cannot be statically analyzed
+ */
+const dynamicCJSModules = new Set(['prop-types', 'react-dom', 'react']);
+
+/**
+ * Bundle am npm module entry path into a single file
+ * @param mod The full path of the module to bundle, including subpackage/path
+ * @param id The imported identifier
+ * @param optimize Whether the bundle should be a minified/optimized bundle, or the default quick non-optimized bundle
+ */
+export const bundleNpmModule = async (
+  mod: string,
+  id: string,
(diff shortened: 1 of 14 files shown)


src/module-server/bundle-npm-module.ts
```diff
@@ -0,0 +1,118 @@
+import type { Plugin, RollupCache } from 'rollup';
```
calebeby, line 1: Nothing in here has changed, it was just moved out of npm-plugin

src/module-server/node-resolve.test.ts
```diff
@@ -0,0 +1,222 @@
...
+  return { resolve };
+};
+
+describe('resolving in node_modules', () => {
+  test('resolves main field', async () => {
+    const fs = await createFs(`
+      ./node_modules/foo/package.json {"main": "entry.js"}
+      ./node_modules/foo/entry.js
+    `);
+    expect(await fs.resolve('foo')).toBe('./node_modules/foo/entry.js');
```
calebeby, line 67: What I've done in the past for something like this is to make real files/folders for each test, but I like this better because it is easier to see the whole test in one place, and because changes are easier to make and will show up better in diffs, etc.

src/module-server/node-resolve.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    const fs = await createFs(`
+      ./folder/package.json {"main": "./cjs.js", "module": "./mjs.js", "exports": "./dont-use.js"}
+      ./folder/mjs.js
+    `);
+    expect(await fs.resolve('./folder')).toBe('./folder/mjs.js');
+  });
+});
+
+test.todo('LOAD_PACKAGE_SELF');
+test.todo('LOAD_PACKAGE_IMPORTS');
```
calebeby, line 222: I left these tests out for now. Package imports are node's new way of creating "aliases" to files in your project, so you can do stuff like `import "#utils` instead of putting large relative paths. I didn't implement support for that yet. And package self refers to being able to import a package in itself (like how in this project we `import "pleasantest"` instead of `import '../dist/...'`). Support for that also isn't implemented yet.

Both are edge cases and I'll get to them eventually, or when someone asks for it

gerardo-rodriguez:
@calebeby I've started reviewing this, I'll finish reviewing it tomorrow. 🔍 👀

src/module-server/node-resolve.test.ts
```diff
@@ -0,0 +1,222 @@
...
+  });
+
+  test('resolves module field over main field and over index.js', async () => {
+    const fs = await createFs(`
+      ./node_modules/foo/package.json {"main": "index.js", "module": "index.esm.js"}
+      ./node_modules/foo/index.esm.js
+      ./node_modules/foo/index.js
+    `);
+    expect(await fs.resolve('foo')).toBe('./node_modules/foo/index.esm.js');
+  });
```
gerardo-rodriguez, line 85: The test name tells me `'resolves module field over main field and over index.js'` but I am not clear how the test assertions confirm that the main field resolves over `index.js`. 

Should the test be renamed? Should there be another test that confirms the main field resolves over `index.js`? 🤔

calebeby: "resolves module field over main field and over index.js" is meant to mean that `module` is resolved with a higher priority than both main and index.js. I can change the "resolves main field" test to also make sure that main has priority over index.js

gerardo-rodriguez: Oh, I see, I misread. 

> I can change the "resolves main field" test to also make sure that main has priority over index.js

Sure! 👍

gerardo-rodriguez (approved):
LGTM! 👍

PR #163, opened 2021-07-15
Make module server configurable

### LINKED ISSUE #140: Make module server configurable
See https://github.com/cloudfour/pleasantest/pull/115

calebeby:
Closes #140 

The code is all ready for review. I didn't update the README yet.

This makes it possible to configure the module server in each test. You could inject additional rollup plugins, etc. An example of this is to add a plugin to load Vue SFC's:

```js
import vuePlugin from 'rollup-plugin-vue'
test(
  'vue component can be imported via rollup-plugin-vue',
  withBrowser(
    {
      moduleServer: {
        plugins: [vuePlugin()],
      },
    },
    async ({ utils, screen }) => {
      await utils.injectHTML('<div id="app"></div>')
      await utils.runJS(`
	    import { createApp } from 'vue'
	    import VueComponent from './vue-component.vue'
	    const app = createApp(VueComponent)
	    app.mount('#app')
	  `)
      const heading = await screen.getByRole('heading')
      await expect(heading).toHaveTextContent('Hiya')
      await expect(heading).toHaveStyle({ color: 'green' })
    },
  ),
)
```

There is also a `configureDefaults` function exported by pleasantest, so you don't have to set the options for each test individually:

```js
import { withBrowser, configureDefaults } from 'pleasantest'

configureDefaults({
  moduleServer: {
    plugins: [vuePlugin()],
  },
})
```

There is also an internal change: File `id`s (used by the rollup plugins) are now absolute paths once they are resolved. This was changed to match Rollup's default behavior, and it fixed a compatibility issue with rollup-plugin-vue. So the ids start out as url's, which are relative to the project root, then they get converted to absolute paths, and run through all the plugins, and at the very end before the module is served they get converted back into paths that are relative to the project root.


src/module-server/transform-imports.ts
```diff
@@ -51,8 +55,32 @@ export const transformImports = async (
...
+    const frame = createCodeFrame(code, line - 1, column);
+    let message = `${colors.red(colors.bold(error.message))}
+
+${colors.red(`${id}:${line}:${column + 1}`)}
+
+${frame}
+`;
+    if (!jsExts.test(id) && !cssExts.test(id))
+      message += `${colors.yellow(
+        'You may need to add transform plugins to handle non-JS input',
```
Paul-Hebert, line 76: Nit: I'm not sure I would understand this error if I ran into it while using Pleasantest. Is there any additional context we could add to make this clearer?

Paul-Hebert (approved):
This is a killer feature @calebeby !

I'm still wrapping my head around this codebase, but these changes seem reasonable :+1: One nitpick about an error message inline.

gerardo-rodriguez (approved):
Super cool feature, great work, @calebeby!

I'm in the same boat as @Paul-Hebert where I don't understand every line of code fully but overall, the code looks good! Pending the suggestion from Paul, I approve!

src/module-server/transform-imports.ts
calebeby: How does this sound: `You may need to add plugins to the module server to handle <extension> files`

calebeby:
Thanks for the reviews @Paul-Hebert and @gerardo-rodriguez! Do you mind taking a look at the README changes that I just pushed as well?

src/module-server/transform-imports.ts
calebeby: This error _only_ appears if the extension is not .js/.cjs/.mjs/.ts/.jsx/.tsx/.css/.sass/.scss/.less/.styl so I don't think it will be very common. I think that adding the file extension would make it more helpful because then people will hopefully understand why the file can't be parsed (i.e. if someone tries to load a .coffee file 😂)

gerardo-rodriguez: > How does this sound: `You may need to add plugins to the module server to handle <extension> files`

I feel this is more helpful, what do you think, @Paul-Hebert? 🙂

Paul-Hebert: Me too!

README.md
```diff
@@ -345,10 +345,28 @@ Call Signatures:
...
 
-`WithBrowserOpts`:
+`WithBrowserOpts` (all properties are optional):
 
 - `headless`: `boolean`, default `true`: Whether to open a headless (not visible) browser. If you use the `withBrowser.headed` chain, that will override the value of `headless`.
 - `device`: Device Object [described here](https://pptr.dev/#?product=Puppeteer&version=v10.1.0&show=api-pageemulateoptions).
+- `moduleServer`: Module Server options object (all properties are optional)
+  - `plugins`: Array of Rollup, Vite, or WMR plugins to add. They will be applied to files imported through [`utils.runJS`](#pleasantestutilsrunjscode-string-promisevoid) or [`utils.loadJS`](#pleasantestutilsloadjsjspath-string-promisevoid).
+  - `envVars`: Object with string keys and string values for environment variables to pass in as `import.meta.env.*` / `process.env.*`
+  - `esbuild`: [`TransformOptions`](https://esbuild.github.io/api/#transform-api) | `false`: Options to pass to esbuild. Set to false to disable esbuild.
```
Paul-Hebert, line 355: Are these just applied when using `utils.runJS`/`utils.loadJS`? Or are they run in other scenarios?

Paul-Hebert:
@calebeby the README info looks good! I left a couple questions inline.

README.md
calebeby: Only using those functions.

Well actually...
- Hypothetically if you used injectHTML and put in some <script> tags I _think_ it would probably serve them from the module server? But this is not documented and not intended to be used
- If you use loadCSS it uses the module server as well. I am not completely sure how user plugins will interact with rollup-plugin-postcss that is there by default, but I don't think user plugins will end up processing those files (I could be wrong)

README.md
```diff
@@ -683,6 +701,10 @@ test(
 );
 ```
 
+todo document configuredefaults
+todo document module server options
+todo explain how module server works
```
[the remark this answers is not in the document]
calebeby: Oops I meant to delete that thanks

README.md
Paul-Hebert: Gotcha. I ask since you specified the `plugins` only apply in those contexts, but it was less clear for me for the `esbuild` options. Maybe it would be good to add it to that bullet as well?

Or, if all of these properties only affect those methods, maybe that should be moved up to the `moduleServer` bullet point?

calebeby: Yeah I think they should be moved up to moduleServer 👍

Issue #175, opened 2021-07-22
LoadJS is missing the source-map-based-error-generation from runJS


calebeby:
I think it is only a matter of extracting the source map -> error code from runJS and sharing it with loadJS

calebeby:
I have this working, just waiting on https://github.com/cloudfour/pleasantest/issues/156 to be merged. Confirmed that it works for runtime errors and syntax errors. Some tests are written, but I will probably write a couple more before PR

PR #190, opened 2021-07-28
Throw module server errors from runJS

### LINKED ISSUE #156: Try to figure out a way to attach errors thrown by module server back to the test code they came from
If I do:

```js
await utils.runJS(`
  console.log('tooManyParens'))
`)
```

then ESBuild throws a parse error, which gets caught and displayed in the console, and the runJS will also throw because the HTTP server returned a 500 for the module.

The problem is that the actual error that ESBuild threw (with the line/col of the syntax error) doesn't get attached to the error thrown by runJS. This is mostly OK, but it makes debugging a little harder because the location of the logged error in the console thrown by ESBuild is not necessarily at the same place as the error thrown by runJS. Also, it means that we can't run error message snapshots for runJS.

## Potential solution #1

One solution that I had experimented with was, when a 500 was thrown, attaching the error message from ESBuild into the response, and changing it to a 200 response, like transpiling the module to: `throw new Error('SyntaxError from ESBuild: location, line/col, etc.')`. The problem with this is that for static imports, the browser doesn't execute the module unless it contains the export names that were imported. Since the original module has syntax errors, we can't parse it to get the export names. I really liked this solution though, and it seems like it is very close to working. Maybe we could use a sketchy regex to detect the export names even if the module is not parseable? (a la [`cjyes`](https://gist.github.com/developit/96de429483bb98927c7cd27c773b0fff`)). Or use a very forgiving parser that works with incorrect syntax like Typescript's?

## Potential solution #2

Another solution would be to create the static module graph in memory in node as a set of promises, and then whenever a module fails to parse/transform/etc, reject all the promises for all the modules above it, bubbling up to the top-level module that gets imported by runJS. The rejection could be detected by runJS (or the top-level module could get awaited?), and then it could incorporate the error message into the output. (I think this won't work for dynamic imports, but we could combine it with solution #1 for dynamic imports?)

## Potential solution #3 

Rewrite all the imports to have a query parameter that is specific to the individual build. Then when an error is thrown, we can reject a promise corresponding to the build ID, which will cause runJS to reject with the error. This would work for dynamic imports, but probably not for dynamic imports which are not static strings

calebeby:
Closes #156

Here's how it looks now!

Screen Shot 2021-07-28 at 3 52 37 PM

## Description

From #156:

> If I do:
> 
> ```js
> await utils.runJS(`
>   console.log('tooManyParens'))
> `)
> ```
> then ESBuild throws a parse error, which gets caught and displayed in the console, and the runJS will also throw because the HTTP server returned a 500 for the module.

There are several problems with this:
- The actual error message that we care about is just logged to the console, so we can't make tests that the error message is correct
- There are two error messages displayed, and the one displayed more prominently is the unimportant one from the browser saying that there was a 500 response for a module

This PR solves this problem, the comment inside the build status tracker file explains how:

> This makes it so that errors thrown by the module server get re-thrown inside of runJS/loadJS.
> This is necessary because the network is between runJS getting called and the module server, so errors do not propagate
> By tracking which runJS/loadJS initiated the request for each file, the module server can know which runJS/loadJS needs to reject.
> When runJS finishes, it will check to see if the buildStatuses map has any errors corresponding to its buildId.
> If there are any errors, it throws the first one.

We keep track of which files were requested by which runJS using build ID's, which are injected into query parameters, i.e.:

`runJS("import './something'")`

`something.ts?build-id=1`
```ts
import {render} from '/@npm/preact?build-id=1'
import {something} from './asdf.js?build-id=1'
```

Each time runJS is called, it gets a new build ID. The build ids propagate all through the module graph (i.e. when it serves `./asdf.js?build-id=1`, the build id query parameter will get attached to every import in _that_ file).

`runJS("import './something-else'")`

`something-else.ts?build-id=2`
```ts
import {render} from '/@npm/preact?build-id=1' // you would expect this to be build-id=2 but it can't be
import {somethingElse} from './asdf2.js?build-id=2'
```

When something gets imported that was already imported in a previous build, the previous buildId gets reused. This is important because it is necessary that the browser returns a cached instance of the module from the module cache, rather than creating a new instance (which will happen if the query parameters change). For many modules, preact included, if there are multiple instances active, there will be bugs. So by reusing the previous build ID, we can fix this problem.

Other things happening in this PR:

The same createCodeFrame code is not repeated all over the place anymore. Now it is in just one place: the ErrorWithLocation class. This makes things a lot easier and more consistent.

I changed node-resolve to throw more useful error messages in more cases instead of undefined. Look at the node-resolve tests to see the specific changes.

Also note: None of the error-handling changes apply to loadJS, only to runJS for now. https://github.com/cloudfour/pleasantest/issues/175 covers making them share the logic so they both get all the benefits of this.

### DIFF
--- a/src/module-server/build-status-tracker.ts
+++ b/src/module-server/build-status-tracker.ts
@@ -1,5 +1,5 @@
 // This makes it so that errors thrown by the module server get re-thrown inside of runJS/loadJS.
-// This is necessary because the network is between runJS getting called and the module server, so errors do not propagate
+// This is necessary because the network is between the runJS call and the module server, so errors do not propagate
 // By tracking which runJS/loadJS initiated the request for each file, the module server can know which runJS/loadJS needs to reject.
 // When runJS finishes, it will check to see if the buildStatuses map has any errors corresponding to its buildId.
 // If there are any errors, it throws the first one.
--- a/src/module-server/build-status-tracker.ts
+++ b/src/module-server/build-status-tracker.ts
@@ -1,0 +1,37 @@
+// This makes it so that errors thrown by the module server get re-thrown inside of runJS/loadJS.
+// This is necessary because the network is between runJS getting called and the module server, so errors do not propagate
+// By tracking which runJS/loadJS initiated the request for each file, the module server can know which runJS/loadJS needs to reject.
+// When runJS finishes, it will check to see if the buildStatuses map has any errors corresponding to its buildId.
+// If there are any errors, it throws the first one.
+
+const buildStatuses = new Map<number, Error[]>();
+
+let buildIds = 0;
+
+/**
+ * Add an error for a specific buildId.
+ * If the build is already finished (resolved),
+ * triggers an uncaught rejection, with the hopes that it will fail the test.
+ */
+export const rejectBuild = (buildId: number, error: Error) => {
+  const statusArray = buildStatuses.get(buildId);
+  if (statusArray) statusArray.push(error);
+  // Uncaught promise rejection!
+  // Hope that Jest will catch it and fail the test, otherwise it is just logged by Node
+  else Promise.reject(error);
+};
+
+export const createBuildStatusTracker = () => {
+  const buildId = ++buildIds;
+  buildStatuses.set(buildId, []);
+  return {
+    buildId,
+    complete() {
+      const status = buildStatuses.get(buildId);
+      // This should never happen
+      if (!status) throw new Error('Build already completed');
+      buildStatuses.delete(buildId);
+      if (status.length > 0) return status;
+    },
+  };
+};
--- a/src/module-server/middleware/js.ts
+++ b/src/module-server/middleware/js.ts
@@ -1,6 +1,10 @@
 import { dirname, posix, relative, resolve, sep } from 'path';
 import type polka from 'polka';
-import type { SourceDescription } from 'rollup';
+import type {
+  PartialResolvedId,
+  ResolveIdResult,
+  SourceDescription,
+} from 'rollup';
 import type { Plugin } from '../plugin';
 import { createPluginContainer } from '../rollup-plugin-container';
 import { promises as fs } from 'fs';
@@ -15,6 +19,8 @@ import * as esbuild from 'esbuild';
 import { createCodeFrame } from 'simple-code-frame';
 import * as colors from 'kolorist';
 import { Console } from 'console';
+import { rejectBuild } from '../build-status-tracker';
+import { ErrorWithLocation } from '../error-with-location';
 
 interface JSMiddlewareOpts {
   root: string;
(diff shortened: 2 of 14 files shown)


src/module-server/middleware/js.ts
```diff
@@ -153,29 +206,18 @@ export const jsMiddleware = ({
...
-      esbuild.transform(code, { loader: 'js' }).catch((error) => {
-        const err = error.errors[0];
-        const { line, column } = err.location;
-        const frame = createCodeFrame(code as string, line - 1, column);
-        const message = `${colors.red(colors.bold(err.text))}
-
-${colors.red(`${id}:${line}:${(column as number) + 1}`)}
-
-${frame}
-`;
```
calebeby, line 171: I deleted this feature. It is possible that I will add it back in the future, but it is definitely an edge-case and I didn't want to slow down _every_ request by 10-300ms because of it.

This would _only_ catch syntax errors before they got to the browsers if all of these are true:
- The user writes invalid code and they don't have the regular esbuild transform enabled, or a user-added plugin turns valid code invalid after esbuild.
- es-module-lexer thinks the end result code is valid (es-module-lexer is kinda a loose parser so this does happen sometimes)

tests/utils/runJS.test.tsx
```diff
@@ -203,13 +234,107 @@ test(
...
 );
 
-test.todo(
-  'if an imported file has a syntax error the location is source mapped',
+const stripAnsi = (input: string) => input.replace(ansiRegex(), '');
+
+const removeLineNumbers = (input: string) => {
+  const lineRegex = /^(\s*>?\s*)(\d+)/gm;
+  const fileRegex = new RegExp(`${process.cwd()}([a-zA-Z/._-]*)[\\d:]*`, 'g');
+  // console.log('remove line numbers', input, input.replace(lineRegex, ''));
```
gerardo-rodriguez, line 242: Can this line be deleted?
```suggestion
```

calebeby: 🤦 yes

src/module-server/build-status-tracker.ts
```diff
@@ -0,0 +1,37 @@
+// This makes it so that errors thrown by the module server get re-thrown inside of runJS/loadJS.
+// This is necessary because the network is between runJS getting called and the module server, so errors do not propagate
```
gerardo-rodriguez, line 2: There is either an extra word or a word missing but I can't figure it out. 😅

calebeby: I _think_ it has all the words I meant for it to have. Maybe it would be clearer like this?

```suggestion
// This is necessary because the network is between the runJS call and the module server, so errors do not propagate
```

gerardo-rodriguez: Ah, got it. I agree with the change! 🙂

gerardo-rodriguez (approved):
Looks good as far as I can tell. 🙃

PR #216, opened 2021-08-17
Improve forgot-await detection by tracking async calls

calebeby:
When @spaceninja and I were working on some tests, we found a bug where the forgot-await detection didn't work:

```js
const buttonEl = await screen.getByRole("button", { name: /count/i });
await expect(buttonEl).toHaveTextContent(/count is: 0/i);
user.click(buttonEl); // Should trigger forgot-await error
await expect(buttonEl).toHaveTextContent(/count is: 3/i);
```

This is because the current implementation usually only works if the call with `await` missing is the last call inside withBrowser, because it depends on Jest's feature of listening for unhandled promise rejections.

This PR reimplements the forgot-await detection. Now it is centralized, and it works differently. It works by keeping track of all async calls that are _started_ inside `withBrowser`. When `withBrowser` finishes, it checks to see if any of the calls did not finish yet. If the developer remembered to use `await` on all the calls, then all the calls should have already finished, so nothing happens. If the developer forgot `await` on one or more of the calls, `withBrowser` can see that, and throw the error.

This does _not_ handle the case where the call that is missing `await` executes fast enough that it finishes before the rest of the things in `withBrowser`.

Review with whitespace changes turned off, otherwise the diff is a lot bigger than it needs to be

### DIFF
--- a/.changeset/chilled-mails-eat.md
+++ b/.changeset/chilled-mails-eat.md
@@ -1,0 +1,5 @@
+---
+'pleasantest': minor
+---
+
+Improve forgot-await detection
--- a/src/async-hooks.ts
+++ b/src/async-hooks.ts
@@ -1,0 +1,61 @@
+/**
+ * Manages asynchronous calls within withBrowser.
+ * If any async calls are "left over" when withBrowser finishes executing,
+ * then that means the user forgot to await the async calls,
+ * so we should throw an error that indicates that.
+ */
+
+import { removeFuncFromStackTrace } from './utils';
+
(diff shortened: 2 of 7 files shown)


gerardo-rodriguez (commented):
Reviewing this now. Gotta love the tests, they are always the most helpful to understand the intent of the changes. 🙂

gerardo-rodriguez (approved):
LGTM! I didn't see anything that obviously jumped out at me. Great work, @calebeby! 🎉

PR #217, opened 2021-08-18
Fix regression with loadJS always throwing

calebeby:
This is what happens when I don't write a test for the core functionality and use `test.todo` 😬 🙈 

Regression introduced in https://github.com/cloudfour/pleasantest/pull/199

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -383,6 +383,7 @@ const createTab = async ({
     const res = await safeEvaluate(
       loadJS,
       `import(${JSON.stringify(url)})
+        .then(mod => {})
         .catch(e => e instanceof Error
           ? { message: e.message, stack: e.stack }
           : e)`,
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -383,6 +383,7 @@ const createTab = async ({
     const res = await safeEvaluate(
       loadJS,
       `import(${JSON.stringify(url)})
+        .then(mod => {})
```
calebeby, line 386: Here is why this works:

On line 396, that function checks `res` to see if it exists, it assumes that it is an error (that is how it works with runJS). If `res` exists, then it throws it.

In the case of loadJS, `res` is the module object that got imported. So I was seeing a confusing error that said `Thrown: {}`. The addition of this line fixes this bug because it discards the imported module so `res` is undefined. The catch block still works to make `res` exist if something was thrown.

gerardo-rodriguez (approved):
Thanks for the explanation, @calebeby! 👍

Issue #176, opened 2021-07-23
Test out resolution with pnpm


calebeby:
Since pnpm uses symlinks, many node resolution implementations have issues with resolving symlinked packages. We should check that our implementation works correctly with node_modules that are symlinked

PR #219, opened 2021-08-19
Improve node_modules resolver

calebeby:
This PR finishes some edge-ish cases with resolution:

1. The resolver should look in node_modules folders other than the main node_modules folder. This is important for if you have multiple versions of a package. Also, it is best not to rely on the hoisting feature of npm/yarn that is not guaranteed.
2. The resolver should support symlinks in node_modules. Now it does.

For #1, here is an example:

```
node_modules
  preact (10.1.0)
  some-library
    preact (8.0.0)
```

Before this PR, PT only looks at `node_modules/preact`. This is bad for two reasons, one is that `node_modules/preact` isn't guaranteed to exist unless we have it as a direct dep, and also, it will always return 10.1.0, even though when importing from some-library, it should return 8.0.0.

This PR fixes that, so it will find the nearest node_modules to the importer that has the requested package.

Both of these changes are necessary for pnpm support. I tested it out by using pnpm with https://github.com/cloudfour/pleasantest-example-modal, linking my local PT in, and making sure the tess pass.

I added tests for #1 but not for #2. #2 can be tested manually:

## Testing pnpm

In this repo:

```
git pull
git checkout good-resolver
npm i
npm run build
npm i -g pnpm
pnpm link --global
```

Clone https://github.com/cloudfour/pleasantest-example-modal
```
pnpm i
pnpm link --global pleasantest
npm run test
```

Closes https://github.com/cloudfour/pleasantest/issues/176

### DIFF
--- a/src/module-server/middleware/js.ts
+++ b/src/module-server/middleware/js.ts
@@ -67,11 +67,12 @@ export const jsMiddleware = async ({
     try {
       // Normalized path starting with slash
       const path = posix.normalize(req.path);
+      const params = new URLSearchParams(req.query as Record<string, string>);
       let id: string;
       let file: string;
       if (path.startsWith('/@npm/')) {
         id = path.slice(1); // Remove leading slash
-        file = ''; // This should never be read
+        file = params.get('resolvedPath') || '';
       } else {
         // Remove leading slash, and convert slashes to os-specific slashes
         const osPath = path.slice(1).split(posix.sep).join(sep);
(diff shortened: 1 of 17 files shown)


src/module-server/middleware/js.ts
```diff
@@ -67,30 +67,30 @@ export const jsMiddleware = async ({
     try {
       // Normalized path starting with slash
       const path = posix.normalize(req.path);
+      const params = new URLSearchParams(req.query as Record<string, string>);
```
Paul-Hebert, line 70: I wanted to flag the warnings above. They don't seem critical and I don't see them as a merge blocker but wanted to make sure you were aware.

(For some reason GitHub wouldn't let me comment directly on them.

Paul-Hebert (approved):
Code looks good and works as advertised :+1: 🚢

src/module-server/middleware/js.ts
calebeby: Thanks 👍

I'm not sure whether at this point there is a good opportunity to factor out part of this function. For now I think I'll leave it as-is but I'll keep my eyes out for refactoring opportunities 👀

Paul-Hebert: Cool, works for me :+1:

PR #234, opened 2021-08-31
Allow functions to be passed to runJS

### LINKED ISSUE #47: There should be a way to pass functions into runJS
Specific use case that we don't have a workflow for:

```js
test(
  'asdf',
  withBrowser(async ({ utils, screen, user }) => {
    const clickHandler = jest.fn()
    await utils.runJS(
      `
      import {h, render} from 'preact'
	  export default (clickHandler) => 
        render(<button onClick={clickHandler}>Click me</button>)
      `,
      [clickHandler],
    )
    expect(clickHandler).toHaveBeenCalledTimes(0)
    const button = await screen.getByRole('button')
    await user.click(button)
    expect(clickHandler).toHaveBeenCalledTimes(1)
  }),
)
```

That example does not work because clickHandler is a function and functions are not serializable over-the-wire to the browser.

This is a common use case because it is used anytime someone wants to test that something fires the right event listeners/callbacks.

It is probably possible to do something where if a function is passed as a parameter to runJS, runJS uses [page.exposeFunction](https://pptr.dev/#?product=Puppeteer&version=v8.0.0&show=api-pageexposefunctionname-puppeteerfunction) or something similar to expose it to the browser. We'll have to make clear in the documentation that the callback support is limited since any values passed to the callbacks or returned from must be serializable, and if the callback returns a value it is a promise rather than a synchronous value. I think these limitations are OK because even with these limitations, the feature is enough to make the above use case with callbacks/event listeners work.

gerardo-rodriguez:
This feature was co-written with @calebeby. This allows functions to be passed to `runJS`. 

See the new test to help understand the intent.

Closes #47

### DIFF
--- a/.changeset/rotten-news-protect.md
+++ b/.changeset/rotten-news-protect.md
@@ -1,0 +1,5 @@
+---
+'pleasantest': minor
+---
+
+Allow functions to be passed to runJS
--- a/src/index.ts
+++ b/src/index.ts
@@ -297,11 +297,30 @@ const createTab = async ({
 
   await page.goto(`http://localhost:${port}`);
 
+  // eslint-disable-next-line @cloudfour/typescript-eslint/ban-types
+  const functionArgs: Function[] = [];
+
   const runJS: PleasantestUtils['runJS'] = (code, args) =>
     asyncHookTracker.addHook(async () => {
+      await page
+        .exposeFunction('pleasantest_callFunction', (id, args) =>
+          functionArgs[id](...args),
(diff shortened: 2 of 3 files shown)


calebeby:
@gerardo-rodriguez Can you add a changeset?

Paul-Hebert (approved):
Looks good, nice work! 🦖 

(I started typing `:tada` and then the t-rex popped up, so you get a 🦖 instead)

PR #236, opened 2021-09-02
Implement minimal accessibility snapshots

calebeby:
I am leaving the feature undocumented until it is no longer experimental


src/accessibility/browser.ts
```diff
@@ -0,0 +1,77 @@
...
+const enum AccessibilityState {
+  /** This element is accessible and its descendents could be accessible */
+  Accessible,
+  /** This element is innaccessible, but its descendents could be accessible */
+  InaccessibleSelf,
+  /** Both this element and its descendents are inaccessible */
+  InaccessibleSelfAndDescendents,
+}
+
+// TODO in PR: what about role="presentation" or role="none"? Should they be excluded?
```
calebeby, line 20: Flagging this for review.

My brain has only parsed this like 20% so I don't really understand it yet. The current implementation in this PR is definitely wrong. I will either make it match the spec in this PR or in a follow-up.

If somebody understands this well (w.r.t. how to treat child elements) feel free to explain it to me!

https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion
> The following elements are not exposed via the accessibility API and user agents MUST NOT include them in the accessibility tree:
> ...
> - Elements with none or presentation as the first role in the role attribute. However, their exclusion is conditional. In addition, the element's descendants and text content are generally included. These exceptions and conditions are documented in the presentation (role) section.


https://www.w3.org/TR/wai-aria-1.2/#presentation

> The intended use is when an element is used to change the look of the page but does not have all the functional, interactive, or structural relevance implied by the element type, or may be used to provide for an accessible fallback in older browsers that do not support WAI-ARIA.
> For any element with a role of presentation and which is not focusable, the user agent MUST NOT expose the implicit native semantics of the element (the role and its states and properties) to accessibility APIs. However, the user agent MUST expose content and descendant elements that do not have an explicit or inherited role of presentation. Thus, the presentation role causes a given element to be treated as having no role or to be removed from the accessibility tree, but does not cause the content contained within the element to be removed from the accessibility tree.

Paul-Hebert: Oh yeah, weird. MDN makes it even more complicated:

> The presentation role is used to remove semantic meaning from an element and any of its related child elements. For example, a table used for layout purposes could have the presentation role applied to the table element to remove any semantic meaning from the table element and any of its table related children elements, such as table headers and table data elements. Non-table related elements should retain their semantic meaning, however.

Happy to chat more about this, though your quotes made me realize I've been misunderstanding `role="presentation"` (and made me grateful for just using `aria-hidden` instead 😅 )

src/accessibility/browser.ts
```diff
@@ -0,0 +1,78 @@
...
+  indenter + text.split('\n').join(`\n${indenter}`);
+
+const enum AccessibilityState {
+  /** This element is accessible and its descendents could be accessible */
+  Accessible,
+  /** This element is innaccessible, but its descendents could be accessible */
+  InaccessibleSelf,
+  /** Both this element and its descendents are inaccessible */
+  InaccessibleSelfAndDescendents,
+}
```
Paul-Hebert, line 18: I'm a little confused by these names. I'm used to thinking about an experience being somewhere on the accessibility spectrum, so it's not obvious to me what it would mean for an element to be fully accessible or inaccessible.

Does accessible in this context mean that the element is in the accessibility tree? I wonder if we could rename these options to make that clearer? (This may just be me. Curious what others think)

src/accessibility/browser.ts
Paul-Hebert: You may want to flag this in our internal #a11y channel or the A11y slack to get some more expertise

src/accessibility/browser.ts
```diff
@@ -0,0 +1,78 @@
...
+const getAccessibilityState = (element: Element): AccessibilityState => {
+  const computedStyle = getComputedStyle(element);
+  if (
+    (element as HTMLElement).hidden ||
+    element.getAttribute('aria-hidden') === 'true' ||
+    computedStyle.display === 'none'
+  )
+    return AccessibilityState.InaccessibleSelfAndDescendents;
+
+  // An element can have visibility: 'hidden' but its descendents can override visibility
```
Paul-Hebert, line 31: Good call! `visibility` is weird...

src/accessibility/browser.ts
```diff
@@ -0,0 +1,78 @@
...
+  Accessible,
+  /** This element is innaccessible, but its descendents could be accessible */
+  InaccessibleSelf,
+  /** Both this element and its descendents are inaccessible */
+  InaccessibleSelfAndDescendents,
+}
+
+// TODO in PR: what about role="presentation" or role="none"? Should they be excluded?
+// The spec says "it depends on the implementation" basically I think
+const getAccessibilityState = (element: Element): AccessibilityState => {
```
Paul-Hebert, line 22: Nit: I wonder if the function name could be changed to make it clearer that it gets the state of a single element? `getElementAccessibility` or something?

src/accessibility/browser.ts
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export const getAccessibilityTree = (
+  element: Element,
+  opts: AccessibilityTreeOptions,
+): string => {
+  const accessibilityState = getAccessibilityState(element);
+  if (accessibilityState === AccessibilityState.InaccessibleSelfAndDescendents)
+    return '';
+  const { includeDescriptions = true, includeText = false } = opts;
```
Paul-Hebert, line 45: TIL you can use object destructuring with default values.

tests/accessibility/getAccessibilityTree.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  'basic use cases',
+  withBrowser(async ({ utils, page }) => {
+    await utils.injectHTML(`
+      <main>
+        <button>
+          <span aria-hidden="true">+</span>
+          <span>Add to cart</span>
+        </button>
+        <h1>hiiii</h1>
+        <div role="button" tabindex="-1">foo &gt bar</div>
```
Paul-Hebert, line 17: TIL you don't need to end char codes with `;`

tests/accessibility/getAccessibilityTree.test.ts
```diff
@@ -0,0 +1,126 @@
...
+    `);
+    expect(await getAccessibilityTree(body)).toMatchInlineSnapshot(`
+      button "click me"
+        ↳ description: "extended description"
+    `);
+    expect(await getAccessibilityTree(body, { includeText: true }))
+      .toMatchInlineSnapshot(`
+        button "click me"
+          ↳ description: "extended description"
+          text "click me"
```
Paul-Hebert, line 58: I wonder if this should be included when `includeText` is true since the text is already displayed in the accessible name? I'm not sure what the right answer is. Curious what you and others think.

tests/accessibility/getAccessibilityTree.test.ts
```diff
@@ -0,0 +1,126 @@
...
+    const body = await page.evaluateHandle<ElementHandle>(() => document.body);
+    expect(await getAccessibilityTree(body)).toMatchInlineSnapshot(`
+      button "A"
+      button "E"
+      button "G"
+      button "H"
+      button "L"
+    `);
+  }),
+);
```
Paul-Hebert, line 105: Would it be worth extending this test to ensure that child elements of hidden element are properly excluded?

(Or not excluded in the case of `<div style="visibility: hidden"><div style="visibility: visible"></div></div>`)

Paul-Hebert (approved):
This is fantastic @calebeby ! Really, really nice work.

This all looks good to me. I left a few nits inline but nothing major. I'd encourage you to get reviews from other team members on this feature as well since they may have other helpful feedback.

tests/accessibility/getAccessibilityTree.test.ts
Paul-Hebert: I thin you probably have it right as-is. Things could get confusing quickly if you start omitting text that's used as a name/description.

src/accessibility/browser.ts
calebeby: > Does accessible in this context mean that the element is in the accessibility tree

Yes. I mean accessible as "can be accessed". Do you have a suggestion of what to name it? Looking through [the spec](https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion) for language to use...

- Excluded/Included
- Hidden

src/accessibility/browser.ts
calebeby: Yep, it's pretty handy!

tests/accessibility/getAccessibilityTree.test.ts
calebeby: Oops maybe the ever-forgiving HTML parser was just being nice to me 😂

tests/accessibility/getAccessibilityTree.test.ts
calebeby: Good question! Ideally I would like it not to be shown. But `dom-accessibility-api` doesn't yet have a way to see which nodes it already used when computing the name/description: https://github.com/eps1lon/dom-accessibility-api/issues/728. I'll open a PR there so we can have that feature.
Also, FWIW chrome's accessibility tree shows the text nodes in cases like this, even if they are already used in computing the name/description.

tests/accessibility/getAccessibilityTree.test.ts
calebeby: I think cases K and L cover that, right?

calebeby: Oh are you talking about the `hidden` attribute?

src/accessibility/browser.ts
Paul-Hebert: I'm not sure... `Excluded`/`Included` seems good. This is also a total nit so if you think what you have now is clearer that's fine by me.

tests/accessibility/getAccessibilityTree.test.ts
Paul-Hebert: Yeah that makes sense. I think Chrome might be doing it right. I'd rather be verbose than remove things in a way that could be confusing (and I could see this becoming confusing with `aria-labelledby` in the mix.)

tests/accessibility/getAccessibilityTree.test.ts
Paul-Hebert: 🤦 Nope, I was reading to quickly. This looks good :+1:

tests/accessibility/getAccessibilityTree.test.ts
calebeby: I think there are some cases where the text is "used up" in computing the name, and the screen reader can't read the text separately from the name. Ideally in those cases we wouldn't show `text`. I'll think about it more to see if there is a consistent way to make this work. (outside of this PR)

calebeby:
cc @gerardo-rodriguez or @spaceninja do you mind taking a look as well?

src/accessibility/browser.ts
calebeby: FWIW I'm planning to research/ask about this and address it in a follow-up PR

gerardo-rodriguez:
> cc @gerardo-rodriguez or @spaceninja do you mind taking a look as well?

I can help review as well! I probably won't be able to get to this until late today or Monday, FYI.

tests/accessibility/getAccessibilityTree.test.ts
Paul-Hebert: Gotcha, that makes sense :+1:

tests/accessibility/getAccessibilityTree.test.ts
```diff
@@ -0,0 +1,126 @@
...
+        <button>J</button>
+      </div>
+      <div style="visibility: hidden">
+        <button>K</button>
+      </div>
+      <div style="visibility: hidden">
+        <button style="visibility: visible">L</button>
+      </div>
+    `);
+    const body = await page.evaluateHandle<ElementHandle>(() => document.body);
```
gerardo-rodriguez, line 96: Are these the same? Is there a preference for consistency?

I noticed this from line ~20: 
```js
const body: ElementHandle = await page.evaluateHandle(() => document.body);
```

src/accessibility/browser.ts
gerardo-rodriguez: A question I have is what is the goal here?

For example, is the goal of the accessibility tree test to validate I purposefully added a `presentation` `role` to an HTML element? If yes, then it seems it'd be helpful to have this information included in the snapshot. 🤔 

Did I make sense?

src/accessibility/index.ts
```diff
@@ -0,0 +1,57 @@
+import type { ElementHandle } from 'puppeteer';
+import { createClientRuntimeServer } from '../module-server/client-runtime-server';
+import { assertElementHandle } from '../utils';
+
+const accessibilityTreeSymbol: unique symbol = Symbol('PT Accessibility Tree');
+
+export interface AccessibilityTreeOptions {
+  /**
+   * Whether the accessibile description of elements should be included in the tree.
```
gerardo-rodriguez, line 9:
```suggestion
   * Whether the accessible description of elements should be included in the tree.
```

tests/accessibility/getAccessibilityTree.test.ts
calebeby: Yeah they end up being the same. I prefer the type argument style since it is tied to Puppeteer's TS API and so if they change the functionality the return type will automatically continue to be correct

src/accessibility/browser.ts
calebeby: @gerardo-rodriguez Good point. Here is how I think about it:

If role="presentation" worked the same way as role="none" (which I don't know if they are but just for the sake of explanation), I would consider which one I was using to be an implementation detail, so they snapshots should be the same regardless of which I'm using. So this makes me reluctant to display a presentation role for `role="presentation"` like we do for other roles:

```
region
  presentation
    heading "hello"
    button "click"
```

It's unclear to me exactly what role="presentation" does to descendent elements. From the MDN page Paul quoted, it seems like descendents that are related to the element with role="presentation" get their roles removed? And maybe other things happen too? Ideally the snapshot would encompass all the changes to the accessibility tree that are described in the spec

src/accessibility/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  );
+
+  return {
+    [accessibilityTreeSymbol]: result,
+    toString: () => result,
+  };
+};
+
+expect.addSnapshotSerializer({
+  serialize: (val, config, indentation, depth, refs, printer) => {
```
gerardo-rodriguez, line 49: I’m assuming this lint warning needs an ignore comment?

src/accessibility/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    options,
+  );
+
+  return {
+    [accessibilityTreeSymbol]: result,
+    toString: () => result,
+  };
+};
+
+expect.addSnapshotSerializer({
```
gerardo-rodriguez, line 48: Can you help me understand what’s going on here? LOL

src/accessibility/index.ts
calebeby: It is a warning so it doesn't fail the linting, it just appears as a warning in the eslint output.

I'll probably disable that rule for this project

src/accessibility/index.ts
calebeby: Jest's inline snapshots have extra quotes for strings:

```js
expect("hello").toMatchInlineSnapshot(`
  "hello"
`)
```

I didn't like the extra quotes for the accessibility tree (it would also escape quotes inside). You can tell Jest how to print certain kinds of objects: https://jestjs.io/docs/expect#expectaddsnapshotserializerserializer

So I made `getAccessibilityTree` return an object and I used `addSnapshotSerializer` to tell Jest how to print the object (without quotes on the outside)

gerardo-rodriguez: Got it. Can you add a brief comment for future devs? Thank you. 🙂

src/accessibility/browser.ts
gerardo-rodriguez: > FWIW I'm planning to research/ask about this and address it in a follow-up PR

I am wondering if we should have a quick Slack call to better understand and/or resolve this with @cloudfour/dev? Or, @calebeby, do you feel your original request/questions have been answered?

Paul-Hebert: I'd be happy to chat. I'm also curious how the presentation role affects its children, and how we should display that info

calebeby: Yeah I think a chat would be good. I could do tomorrow afternoon anytime?

Paul-Hebert: Same :+1:

calebeby:
@gerardo-rodriguez @Paul-Hebert I think I've gotten to all the feedback, except for the role="presentation" conversation which can be a follow up PR

Paul-Hebert:
>@gerardo-rodriguez @Paul-Hebert I think I've gotten to all the feedback, except for the role="presentation" conversation which can be a follow up PR

Cool, my approval still stands. This is looking great 🙂

gerardo-rodriguez (approved):
LGTM, excellent work, @calebeby! 🎉

PR #251, opened 2021-09-13
Handle `childrenArePresentational`and change `includeText` default to true

Paul-Hebert:
# Overview

This PR makes a couple of changes to the `experimentalGetAccessibilityTree` function:

## `childrenArePresentational`

Some roles have a [`childrenArePresentational` attribute](https://www.w3.org/TR/wai-aria-1.1/#childrenArePresentational) set to true. That means:

> The DOM descendants are presentational. User agents SHOULD NOT expose descendants of this element through the platform accessibility API. If user agents do not hide the descendant nodes, some information may be read twice.

For us, that means we should not expose child elements as elements or text nodes within the main element. The text of these elements may still be used in the main element's accessible name. 

We did this by pulling node information from the [aria-query](https://github.com/A11yance/aria-query) library to determine the relevant roles. Then we check whether an element's role has the `childrenArePresentational` attribute set to true, before iterating over its children.

## `includeText` default value is now `true`

The change above reduced the amount of duplicate output when `includeText` is set to `true` so we made that the new default.


src/accessibility/browser.ts
```diff
@@ -8,6 +8,24 @@ import type { AccessibilityTreeOptions } from '.';
 const indent = (text: string, indenter = '  ') =>
   indenter + text.split('\n').join(`\n${indenter}`);
 
+const rolesWithChildrenPresentation = new Set([
```
Paul-Hebert, line 11:
```suggestion
// This data was extracted from the aria-query library
//  https://github.com/A11yance/aria-query/blob/main/scripts/roles.json
const rolesWithChildrenPresentation = new Set([
```

Paul-Hebert:
@calebeby I'm a bit stumped on why CI is failing for some versions of Node. It doesn't seem related to my changes. Any ideas?

calebeby:
@Paul-Hebert the build is failing because `src/accessibility/index.js` exists, and it is importing from there instead of from `src/accessibility/index.ts`

Paul-Hebert:
>@Paul-Hebert the build is failing because src/accessibility/index.js exists, and it is importing from there instead of from src/accessibility/index.ts

Oh, D'oh! I forgot to delete that after using it as a node script. Good catch :+1:

Paul-Hebert:
@calebeby it looks like for some reason not all of our tests were being run locally (maybe because of the index.js issue?)

I reran the tests and the `includeText` change ended up being more verbose in some of our other tests: https://github.com/cloudfour/pleasantest/pull/251/commits/75a8d1edbfc58e44d4300344b691f4e7233cbddd

Does that change your feelings at all about changing the default?

calebeby:
Thanks for updating that. Even with the changes I don't think it is too verbose.

calebeby:
@gerardo-rodriguez or @spaceninja Can you review this? I was involved in writing it so someone other than Paul/I should review it

spaceninja (approved):
The code changes all look reasonable to me.

I assume for testing this PR, we just trust that the checks pass, so it's good?

Paul-Hebert:
>I assume for testing this PR, we just trust that the checks pass, so it's good?

Yeah I think so, unless @calebeby has other thoughts?

calebeby:
> I assume for testing this PR, we just trust that the checks pass, so it's good?

:+1:

PR #248, opened 2021-09-10
Target size feature

gerardo-rodriguez:
Closes #72

PR #290, opened 2021-11-04
Fix stack traces handling when calling user.* methods

calebeby:
This fixes a bug that Gerardo and I found while implementing #248:
In #216, there is a regression in stack traces when an error is thrown from `user.*` methods.

For example, if you try to click an element that is invisible:

```js
await user.click(invisibleButton)
```

Then the correct/ideal behavior would be to show an error message with that line as the first line of the stack.
Since the regression in #216, the stack trace is missing entirely.

This PR adds it back, and adds a test to make sure that it stays there.

### DIFF
--- a/.changeset/two-dingos-develop.md
+++ b/.changeset/two-dingos-develop.md
@@ -1,0 +1,5 @@
+---
+'pleasantest': patch
+---
+
+Fix stack frames handling when calling user.\* methods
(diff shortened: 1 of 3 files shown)


calebeby:
Uh oh my solution works locally but not in CI 🤔

Paul-Hebert (approved):
LGTM!

PR #313, opened 2021-11-29
Accessibility tree snapshots: Handle role="presentation"/role="none"

calebeby:
This handles elements with `role="presentation"`/`role="none"` in the same way that browsers/screen readers do (as best as I can tell).

The complexity is that when a parent element has `role="presentation"`/`role="none"`, any of its accessibility tree children (could be descendents in the DOM tree) which are "required owned elements" also need to be handled as role="presentation".

I used [this resource](https://www.digitala11y.com/presentation-role/) to try to understand how to implement this correctly. It was much more legible to me than the spec.


gerardo-rodriguez:
> I used [this resource](https://www.digitala11y.com/presentation-role/) to try to understand how to implement this correctly. It was much more legible to me than the spec.

Good find! This was definitely easier for me to understand also. 😅

tests/accessibility/getAccessibilityTree.test.ts
```diff
@@ -139,6 +139,76 @@ test(
...
+test(
+  'role="presentation"',
+  withBrowser(async ({ utils, page }) => {
+    const body = (await page.$('body'))!;
+    await utils.injectHTML(`<h1 role="presentation">Sample Content</h1>`);
+    // Role="presentation" and role="none" are equivalent
+    // They make it as if the outer element wasn't there.
+    expect(await getAccessibilityTree(body)).toMatchInlineSnapshot(
+      `text "Sample Content"`,
+    );
```
gerardo-rodriguez, line 151: These comments are very helpful, thank you, @calebeby! 🙂 

Does it make sense to add a link to the [documentation you found](https://www.digitala11y.com/presentation-role/) here also?

gerardo-rodriguez: Or maybe above it?

```js
// See: https://www.digitala11y.com/presentation-role/
test('role="presentation"', withBrowser() => {});
```

tests/accessibility/getAccessibilityTree.test.ts
```diff
@@ -139,6 +139,76 @@ test(
...
+    await utils.injectHTML(`<h1 role="none">Sample Content</h1>`);
+    // Role="presentation" and role="none" are equivalent
+    // They make it as if the outer element wasn't there.
+    expect(await getAccessibilityTree(body)).toMatchInlineSnapshot(
+      `text "Sample Content"`,
+    );
+
+    // The li (role=listitem) children are required owned elements of the ul (role=list)
+    // so when the list is set to role=presentation, the required owned elements are too
+    // The third li has a different role so it is not a required owned element of the list
```
gerardo-rodriguez, line 161: This read a bit funny. I'm not sure my suggestion is better. 🤔 
```suggestion
    // The li (role=listitem) children are required owned elements of the ul (role=list)
    // When the list is set to role=presentation, the required owned elements are also
    // The third li has a different role so it is not a required owned element of the list
```

gerardo-rodriguez: > // The third li has a different role so it is not a required owned element of the list

@calebeby I'm trying to find where this expectation is described in the [document you shared](https://www.digitala11y.com/presentation-role/) but I can't find it. Can you direct me to where this is documented? I simply want to confirm/understand. Thanks! 😉

src/accessibility/rollup.config.js
```diff
@@ -2,14 +2,44 @@ import babel from '@rollup/plugin-babel';
...
+  resolveId: (id) => (id === 'generated:requiredOwnedElements' ? id : null),
+  load(id) {
+    if (id !== 'generated:requiredOwnedElements') return;
+    let output = 'export default new Map([\n';
+    for (const [role, data] of r.roles) {
+      if (data.requiredOwnedElements.length > 0) {
+        // Some of the sub-arrays have two items,
+        // which means that the outer element owns elements
+        // which have a role within another element with another role
+        // Our implementation doesn't support this yet.
```
gerardo-rodriguez, line 26: Is there a GH issue we could link to here?

gerardo-rodriguez (commented):
I've given this a first pass review and left some questions, thanks @calebeby, great work! 🙂

tests/accessibility/getAccessibilityTree.test.ts
calebeby: In the spec it says that the parent role="presentation" should only cascade to accessibility tree children which are required owned elements of the parent element and which have no explicit role defined.

> When an explicit or inherited role of presentation is applied to an element with the implicit semantic of a WAI-ARIA role that has required owned elements, in addition to the element with the explicit role of presentation, the user agent MUST apply an inherited role of presentation to any owned elements that do not have an explicit role defined. Also, when an explicit or inherited role of presentation is applied to a host language element which has required children as defined by the host language specification, in addition to the element with the explicit role of presentation, the user agent MUST apply an inherited role of presentation to any required children that do not have an explicit role defined.

I'm realizing that I should probably add to this test a check for an explicit role which is the same as the implicit role `<li role="listitem">` (I think the correct behavior is for the role="presentation") to not cascade in that case.

calebeby: Confirmed this in Chrome/FF:

```html
<ul role="presentation">
	  <li>Sample Content</li>
	  <li>More Sample Content</li>
	  <li role="listitem">Hi</li>
</ul>
```

When the `<ul>` has role="presentation":

Screen Shot 2021-12-07 at 3 29 07 PM

When the parent has no explicit role:

Screen Shot 2021-12-07 at 3 31 23 PM

calebeby: I pushed a fix and a test to handle this edge case where the explicit role set on a descendant makes it a required owned element, but the presence of the explicitly set role means that the role="presentation" on the parent does not cascade to it.

src/accessibility/rollup.config.js
calebeby: I think it would be difficult to come up with a non-contrived example for this. I think we can open an issue if we run into a scenario where this divergence from the spec actually happens, since then we'll have a real-world example. I think it is unlikely for us to run into a scenario where this difference matters.

tests/accessibility/getAccessibilityTree.test.ts
calebeby: @gerardo-rodriguez re: your initial suggested change in the comment: I rephrased that sentence so I think it reads better now. Let me know what you think!

calebeby:
@gerardo-rodriguez Should be ready for another review pass!

gerardo-rodriguez (approved):
Awesome work, @calebeby! 🎉

PR #327, opened 2021-12-07
Add test for svelte and improve error message when plugin is missing

calebeby:
@Paul-Hebert asked whether Pleasantest supported Svelte and I realized I hadn't actually tested it out. This PR adds a test to make sure it works (it did, without changes).

I also added a suggestion for error messages when there are syntax errors while parsing a non-JS file that is imported into JS (i.e. it will suggest adding a plugin for parsing .vue or .svelte files if you import them and there is a parse error after all user plugins have run)


tests/utils/svelte-component.svelte
```diff
@@ -0,0 +1,20 @@
...
+	Clicked {count} {count === 1 ? 'time' : 'times'}
+</button>
+
+<style lang="scss">
+$bg: red;
+
+button {
+  background: $bg;
+}
+</style>
```
gerardo-rodriguez, line 20: Nit, can you adjust the tabbing in this file, it's all over the place. 😅

gerardo-rodriguez (approved):
One nit comment, otherwise, LGTM! 👍

tests/utils/svelte-component.svelte
calebeby: Oh yeah good point, this is what I get for copy-pasting svelte examples from a bunch of places 😂

Paul-Hebert:
This is awesome, thanks @calebeby !

PR #343, opened 2021-12-21
Mark getAccessibilityTree as stable and document it

calebeby:
After this is merged I will create a minor release.


Paul-Hebert (approved):
This looks great! Fantastic job on the explanatory docs.

PR #344, opened 2021-12-22
Allow passing a Page instead of an ElementHandle to getAccessibilityTree

calebeby:
## Overview

After using `getAccessibilityTree` a bit I realized that it is very common to want to do `getAccessibilityTree` on the whole document, so I made a shortcut to get a reference to the body element:

```js
const body = await screen.container
```

This is equivalent to:

```ts
const body = await page.evaluateHandle<ElementHandle>(() => document.body)
```

Also, it can be used to get a reference to the container element used after calling `within(el)` to get queries scoped to a specific element:

```ts
const elQueries = within(el)
const container = await elQueries.container
```

In this example, `container` and `el` are both `ElementHandle`s that point to the same element.

## Alternate solutions

- Other names for `.container`
  - `root`
  - `el`
  - `element`
- Make `getAccessibilityTree` default to the document if no element is passed
  ```js
  expect(await getAccessibilityTree(/* implied body */)).toMatchInlineSnapshot()
  ```
- Add a parameter that is passed to the `withBrowser` callback
  ```js
  withBrowser(async ({ getBody }) => {
    const body = await getBody()
  })
  ```

I don't know which of these solutions would be best (it is also possible to implement multiple of them). Let me know if y'all have thoughts on which would be best, it would be pretty easy to implement any of them.

### DIFF
--- a/.changeset/tall-spies-type.md
+++ b/.changeset/tall-spies-type.md
@@ -2,21 +2,6 @@
 'pleasantest': minor
 ---
 
-Add `screen.container` property
+Allow passing `page` instead of an `ElementHandle` to `getAccessibilityTree`.
 
-This is an easy way to get a reference to document.body:
-
-```js
-const body = await screen.container;
-```
-
-Anytime you access the container property, it will be an up-to-date reference to the latest `body`.
-
-This can also be used on queries bound to elements other than `screen`, to retrieve the container the queries were bound to:
-
-```js
-const elQueries = within(el);
-const container = await elQueries.container;
-```
-
-In this example, `container` and `el` are both `ElementHandle`s that point to the same element.
+If `page` is passed, the accessibility tree will be of the root `html` element.
--- a/.changeset/tall-spies-type.md
+++ b/.changeset/tall-spies-type.md
@@ -1,0 +1,22 @@
+---
+'pleasantest': minor
+---
+
+Add `screen.container` property
+
+This is an easy way to get a reference to document.body:
+
+```js
+const body = await screen.container;
+```
+
+Anytime you access the container property, it will be an up-to-date reference to the latest `body`.
+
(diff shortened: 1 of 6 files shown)


Paul-Hebert (approved):
This code looks good to me and it makes sense to have a quick way to grab the body.

I'm less clear on the use case of `queryEl.container`. In the examples you documented this variable is always matching an existing variable. Could you share more context for when you'd need this? It might be nice to add that context to the docs.

>Make getAccessibilityTree default to the document if no element is passed 

This alternative solution makes a lot of sense to me!

calebeby:
> I'm less clear on the use case of `queryEl.container` ...

Me too 😂 . I mostly put it there so that it parallels the `screen.container` property since `screen` and `within(body)` are supposed to be the same thing.

Here is a somewhat contrived use case:

```
const heading = within(await screen.getByRole('heading'))

await heading.getByText('...') // etc

initHeading(heading.container);
```

Since the `screen.getByRole('heading')` query is passed directly in to `within`, it doesn't have its own variable, so the `.container` property is helpful here. But it could just as easily use a pair of variables, `heading` and `headingEl` to access the queries and the container.

> Make getAccessibilityTree default to the document if no element is passed

Yeah I think this solution probably is better. I'll implement that and we can talk about removing `.container`

Paul-Hebert:
Sounds good :+1: If we end up leaving in `queryEl.container` we may want to update the docs to a use case similar to the one you mentioned above.

calebeby:
@Paul-Hebert I took out screen.container. Maybe we can re-add it if we have a good use case for it.

Now getAccessibilityTree accepts a Page as the first parameter, and if that is passed it will print the accessibility snapshot of the entire document (from the `<html>` element, not `<body>`).

Paul-Hebert (approved):
LGTM!

PR #345, opened 2021-12-22
Normalize whitespace in getAccessibilityTree

calebeby:
Now anytime there is contiguous whitespace in text strings it is collapsed into a single space. This matches the behavior of browser accessibility trees.

This is a breaking change because it changes the `getAccessibilityTree` output, and may break your snapshots. Update your snapshots with Jest and review the changes.

Since it is a breaking change I will hold off on merging it


tests/accessibility/getAccessibilityTree.test.ts
```diff
@@ -58,24 +56,23 @@ test(
         ↳ description: "extended description"
       text "extended description"
     `);
-    expect(await getAccessibilityTree(body, { includeText: true }))
```
calebeby, line 61: Since we changed the default value of includeText to true, this snapshot and the one above are the same. I changed the `includeText` property here so we test both cases.

Paul-Hebert (approved):
LGTM!

PR #369, opened 2022-01-11
Add waitFor

calebeby:
This is something in dom testing library that I had skipped but it seems like there are a fair number of use cases where it is useful

https://testing-library.com/docs/dom-testing-library/api-async/#waitfor

### DIFF
--- a/src/pptr-testing-library.ts
+++ b/src/pptr-testing-library.ts
@@ -4,7 +4,7 @@ import {
   printColorsInErrorMessages,
   removeFuncFromStackTrace,
 } from './utils';
-import type { ElementHandle, JSHandle } from 'puppeteer';
+import type { ElementHandle, JSHandle, Page } from 'puppeteer';
 import { createClientRuntimeServer } from './module-server/client-runtime-server';
 import type { AsyncHookTracker } from './async-hooks';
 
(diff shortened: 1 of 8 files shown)


tests/wait-for.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  'Basic case',
+  withBrowser(async ({ utils, page, waitFor }) => {
+    await utils.injectHTML('<h1></h1>');
+    await utils.runJS(`
+      setTimeout(() => {
+        document.write('<h2>Hi</h2>')
+      }, 100)
+    `);
+    // At first the element should not be there
+    // Because it waits 10ms to add it
```
Paul-Hebert, line 14:
```suggestion
    // Because it waits 100ms to add it
```

tests/wait-for.test.ts
```diff
@@ -0,0 +1,83 @@
...
+                         ^
+      -------------------------------------------------------
+      dist/cjs/index.cjs"
+    `);
+
+    // Allows customizing error message using onTimeout
+    const error3 = await waitFor(() => new Promise<never>(() => {}), {
+      timeout: 10,
+      onTimeout: (err) => {
+        err.message += '\nCaleb wuz here';
```
Paul-Hebert, line 67: 😆

src/pptr-testing-library.ts
```diff
@@ -216,3 +216,95 @@ export const getQueriesForElement = (
...
+  /**
+   * The amount of time (milliseconds) that will pass before waitFor "gives up" and throws whatever the callback threw.
+   * Default: 1000ms
+   */
+  timeout?: number;
+  /**
+   * The maximum amount of time (milliseconds) that will pass between each run of the callback.
+   * If the MutationObserver notices a DOM change before this interval triggers,
+   * the callback will run again immediately.
+   * Default: 50ms
```
Paul-Hebert, line 239: Are we actually setting these defaults  somewhere or are these just the defaults from testing library?

If they're from testing library are we worried at all that are docs could be wrong if they change those defaults? Maybe we should set these defaults to make sure the docs and implementation stay in sync?

Paul-Hebert (approved):
Looks great! Left a few comments/questions inline.

src/pptr-testing-library.ts
calebeby: These are the defaults from testing library.
Whenever there are DTL updates, I check through their changes to see if there is anything on our end we need to update. I agree that I don't want our docs to diverge from the functionality, but I also don't want our functionality to diverge from DTL's functionality. Whether we set the defaults in code here or not, we'd still have to keep an eye out to prevent it from diverging

PR #403, opened 2022-02-08
Expose `accessibilityTreeSnapshotSerializer`

calebeby:
See the [changeset](https://github.com/cloudfour/pleasantest/blob/accessibilityTreeSnapshotSerializer/.changeset/spicy-peas-help.md) for a description.

I did not document this workflow in the README yet since I want to give the idea some time to settle and get used on projects/etc before we add it to the README as a solid workflow.

### DIFF
--- a/.changeset/spicy-peas-help.md
+++ b/.changeset/spicy-peas-help.md
@@ -1,0 +1,52 @@
+---
+'pleasantest': minor
+---
+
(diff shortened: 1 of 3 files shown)


Paul-Hebert (approved):
LGTM! We'll want to be sure to write very clear setup instructions when this is formally documented.

PR #446, opened 2022-03-29
Use document.title as fallback accessible name for html root element

calebeby:
I could not find this in the spec, but it is consistent in Chrome's devtools accessibility inspector, Firefox's, and also in VO in safari:

The HTML root element gets the role of `document` (this _is_ in the spec)
That element gets an implicit name corresponding to document.title (I could not find this in the spec).

Here is where I looked:
- https://www.w3.org/TR/wai-aria-1.1/#document
- https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_te
- https://www.w3.org/TR/html-aria/#el-html

### DIFF
--- a/.changeset/popular-beans-roll.md
+++ b/.changeset/popular-beans-roll.md
@@ -1,0 +1,5 @@
+---
+'pleasantest': major
+---
+
+Use document.title as fallback implicit accessible name for html root element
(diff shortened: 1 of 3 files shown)


spaceninja (approved):
👍🏻

PR #451, opened 2022-03-31
Add `aria-expanded` support to `getAccessibilityTree` and fix handling for `<details>`/`<summary>`

calebeby:


tests/accessibility/getAccessibilityTree.test.ts
```diff
@@ -281,3 +281,68 @@ test(
...
+    `);
+
+    const toggle = await screen.getByText(/click me!/i);
+    await user.click(toggle);
+
+    // After toggling it should be expanded
+    expect(await getAccessibilityTree(page)).toMatchInlineSnapshot(`
+      document
+        group
+          button "Click me! Tags in summary do not preserve their semantic meaning" (expanded=true) (focused)
```
calebeby, line 313: This behavior is according to spec (the `<h1>` inside of `<summary>` loses its semantic meaning). But, the behavior does not match voiceover in safari, or chrome or firefox's accessibility tree.

According to the spec:
- [`<summary>` implicitly gets the `button` role](https://www.w3.org/TR/html-aria/#el-summary)
- [The `button` role is "children presentational"](https://www.w3.org/TR/wai-aria-1.1/#button)
- ["children presentational" elements should not directly expose their children in the accessibility tree](https://www.w3.org/TR/wai-aria-1.1/#childrenArePresentational)

It would be more difficult for us to diverge from the spec and implement it the way browsers do. Chrome accessibility tree, firefox accessibility tree, and safari voiceover all chose not to use the `button` role for `<summary>` elements. We could add a special override here but I would rather not.

calebeby: Here are what browsers use for `<summary>` elements:
- Chrome devtools accessibility tree: `DisclosureTriangle`
- Firefox devtools accessibility tree: `summary`
- Safari VO: `summary, group`

spaceninja: I think it's reasonable to leave this matching the spec. Might be good to add a "known issues" section in the README, though, outlining things like this that might otherwise trigger bug reports.

PR #494, opened 2022-06-02
Add toPassAxeTests feature

calebeby:
This assertion, based on [`jest-puppeteer-axe`](https://github.com/WordPress/gutenberg/tree/3b2eccc289cfc90bd99252b12fc4c6e470ce4c04/packages/jest-puppeteer-axe), allows you to check a page using the [axe accessibility linter](https://github.com/dequelabs/axe-core).

The output is a little nicer than jest-puppeteer-axe's because it shows the markup of the elements which are breaking axe rules (as opposed to a non-unique CSS selector), so it is easier to figure out which elements need to be fixed. In the browser, you can click on the markup to see the relevant elements in the devtools elements panel.

[Documentation](https://github.com/cloudfour/pleasantest/tree/to-pass-axe-tests#expectpagetopassaxetestsopts-topassaxetestsopts)

```js
test(
  'Axe tests',
  withBrowser(async ({ utils, page }) => {
    await utils.injectHTML(`
      <h1>Some html</h1>
    `);
    await expect(page).toPassAxeTests();
  }),
);
```

Here's what the output looks like in the CLI:

Screen Shot 2022-06-02 at 8 45 29 AM

Here's what it looks like in the browser:

Screen Shot 2022-06-02 at 8 45 53 AM

## Testing

Run the tests in watch mode:
```
git checkout to-pass-axe-tests
npm i
npm run build
npm run test:watch topassaxe
```

Note: the HTML syntax highlighting in the CLI output isn't shown when Pleasantest runs in its own test suite. This is so that the test snapshots are still readable and don't get covered with escape codes for coloring.

Test the browser output by changing [`tests/accessibility/toPassAxeTests.test.ts`](https://github.com/cloudfour/pleasantest/blob/to-pass-axe-tests/tests/accessibility/toPassAxeTests.test.ts) to use `withBrowser.headed` instead of `withBrowser`. Then it should open a browser for the test, and close it once it passes. To make the test fail so you can see the failure message in the browser, under the `utils.injectHTML(...)` line add a new line:
```js
await expect(page).toPassAxeTests()
```

### DIFF
--- a/src/accessibility/browser.ts
+++ b/src/accessibility/browser.ts
@@ -12,7 +12,7 @@ export * as colors from 'kolorist';
 export { printElement } from '../serialize';
 
 // We haver to tell kolorist to print the colors
-// beacuse by default it won't since we are in the browser
+// because by default it won't since we are in the browser
 // (the colored message gets sent to node to be printed)
 colors.options.enabled = true;
 colors.options.supportLevel = 1;
--- a/src/accessibility/browser.ts
+++ b/src/accessibility/browser.ts
@@ -6,6 +6,16 @@ import {
 // @ts-expect-error This is a fake file that triggers a rollup plugin
 import requiredOwnedElementsMap from 'generated:requiredOwnedElements';
 import type { AccessibilityTreeOptions } from '.';
+import * as colors from 'kolorist';
+
+export * as colors from 'kolorist';
+export { printElement } from '../serialize';
+
+// We haver to tell kolorist to print the colors
+// beacuse by default it won't since we are in the browser
+// (the colored message gets sent to node to be printed)
+colors.options.enabled = true;
+colors.options.supportLevel = 1;
 
 const indent = (text: string, indenter = '  ') =>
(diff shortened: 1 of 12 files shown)


src/accessibility/browser.ts
```diff
@@ -6,6 +6,16 @@ import {
 // @ts-expect-error This is a fake file that triggers a rollup plugin
 import requiredOwnedElementsMap from 'generated:requiredOwnedElements';
 import type { AccessibilityTreeOptions } from '.';
+import * as colors from 'kolorist';
+
+export * as colors from 'kolorist';
+export { printElement } from '../serialize';
+
+// We haver to tell kolorist to print the colors
+// beacuse by default it won't since we are in the browser
```
Paul-Hebert, line 15:
```suggestion
// because by default it won't since we are in the browser
```

Paul-Hebert (approved):
This is awesome! Code looks good to me besides one nit.

PR #510, opened 2022-06-16
Change imports to real ESM

calebeby:
This comes along with the eslint plugin update: all the imports are sorted and grouped, node internal imports are prefixed with `node:`, and all imports are changed to use `.js` extensions instead of omitting extensions.

I enabled the TS settings to follow Node's ESM resolution. Many npm packages don't yet correctly support this so I had to create a `types.d.ts` file to adjust some of their types, and override some of the types resolutions in `tsconfig.json`.

It is easiest to review the commits separately; the 2nd commit is all automated changes so it doesn't need careful review.

### DIFF
--- a/examples/menu/index.test.ts
+++ b/examples/menu/index.test.ts
@@ -1,8 +1,9 @@
-import type { PleasantestUtils } from 'pleasantest';
-import { getAccessibilityTree, withBrowser, devices } from 'pleasantest';
+import * as path from 'node:path';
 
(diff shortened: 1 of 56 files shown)


Paul-Hebert (approved):
@calebeby these changes all look good to me. Should I be doing any testing steps on this or are we confident it works from our CI checks?

calebeby:
The tests are pretty solid!

Issue #522, opened 2022-07-05
Return exports from runJS, and pass in args via import.meta.pleasantestArgs


calebeby:
```ts
const exp = await runJS(`
  export const s = 54
`)
exp.s // => JSHandle
exp.s.jsonValue() // => 54
```

```ts
const exp = await runJS(`
  const [hello] = import.meta.pleasantestArgs
  console.log(hello) // 'hello'
`, ['hello'])
```

This is a breaking change, `import.meta.pleasantestArgs` replaces the default implicitly-called function export.

calebeby:
After a slack discussion, we have decided:

1. The code inside the runJS block will rerun every time runJS is called, even if the code inside is the same.
2. Exported values from the module passed into runJS will be returned as an object by runJS. The exported values will be available as JSHandles.

PR #535, opened 2022-07-27
Change runJS parameter passing and return value

calebeby:
Closes https://github.com/cloudfour/pleasantest/issues/522

There are only like ~15 lines of code changes here! I added a lot of comments, tests, and documentation.

Recommend reviewing with whitespace changes hidden.

Overview of the changes in the changesets:

> 
> Values exported from `runJS` are now available in Node.
> 
> For example:
> 
> ```js
> test(
>   'receiving exported values from runJS',
>   withBrowser(async ({ utils }) => {
>     // Each export is available in the returned object.
>     // Each export is wrapped in a JSHandle, meaning that it points to an in-browser object
>     const { focusTarget, favoriteNumber } = await utils.runJS(`
>       export const focusTarget = document.activeElement
>       export const favoriteNumber = 20
>     `);
> 
>     // Serializable JSHandles can be unwrapped using JSONValue:
>     console.log(await favoriteNumber.jsonValue()); // Logs "20"
> 
>     // A JSHandle<Element>, or ElementHandle is not serializable
>     // But we can pass it back into the browser to use it (it will be unwrapped in the browser):
> 
>     await utils.runJS(
>       `
>       // The import.meta.pleasantestArgs context object receives the parameters passed in below
>       const [focusTarget] = import.meta.pleasantestArgs;
>       console.log(focusTarget) // Logs the element in the browser
>       `,
>       // Passing the JSHandle in here passes it into the browser (unwrapped) in import.meta.pleasantestArgs
>       [focusTarget],
>     );
>   }),
> );
> ```
> 
> We've also introduced a utility function to make it easier to call `JSHandle`s that point to functions, `makeCallableJSHandle`. This function takes a `JSHandle<Function>` and returns a node function that calls the corresponding browser function, passing along the parameters, and returning the return value wrapped in `Promise<JSHandle<T>>`:
> 
> ```js
> // new import:
> import { makeCallableJSHandle } from 'pleasantest';
> 
> test(
>   'calling functions with makeCallableJSHandle',
>   withBrowser(async ({ utils }) => {
>     const { displayFavoriteNumber } = await utils.runJS(`
>       export const displayFavoriteNumber = (number) => {
>         document.querySelector('.output').innerHTML = "Favorite number is: " + number
>       }
>     `);
> 
>     // displayFavoriteNumber is a JSHandle<Function>
>     // (a pointer to a function in the browser)
>     // so we cannot call it directly, so we wrap it in a node function first:
> 
>     const displayFavoriteNumberNode = makeCallableJSHandle(
>       displayFavoriteNumber,
>     );
> 
>     // Note the added `await`.
>     // Even though the original function was not async, the wrapped function is.
>     // This is needed because the wrapped function needs to asynchronously communicate with the browser.
>     await displayFavoriteNumberNode(42);
>   }),
> );
> ```
> 
> For TypeScript users, `runJS` now accepts a new optional type parameter, to specify the exported types of the in-browser module that is passed in. The default value for this parameter is `Record<string, unknown>` (an object with string properties and unknown values). Note that this type does not include `JSHandles`, those are wrapped in the return type from `runJS` automatically.
> 
> Using the first example, the optional type would be:
> 
> ```ts
> test(
>   'receiving exported values from runJS',
>   withBrowser(async ({ utils }) => {
>     const { focusTarget, favoriteNumber } = await utils.runJS<{
>       focusTarget: Element;
>       favoriteNumber: number;
>     }>(`
>       export const focusTarget = document.activeElement
>       export const favoriteNumber = 20
>     `);
>   }),
> );
> ```
> 
> Now `focusTarget` automatically has the type `JSHandle<Element>` and `favoriteNumber` automatically has the type `JSHandle<number>`. Without passing in the type parameter to `runJS`, their types would both be `JSHandle<unknown>`.

---

> The way that `runJS` receives parameters in the browser has changed. Now, parameters are available as `import.meta.pleasantestArgs` instead of through an automatically-called default export.
> 
> For example, code that used to work like this:
> 
> ```js
> test(
>   'old version of runJS parameters',
>   withBrowser(async ({ utils }) => {
>     // Pass a variable from node to the browser
>     const url = isDev ? 'dev.example.com' : 'prod.example.com';
> 
>     await utils.runJS(
>       `
>       // Parameters get passed into the default-export function, which is called automatically
>       export default (url) => {
>         console.log(url)
>       }
>       `,
>       // array of parameters passed here
>       [url],
>     );
>   }),
> );
> ```
> 
> Now should be written like this:
> 
> ```js
> test(
>   'new version of runJS parameters',
>   withBrowser(async ({ utils }) => {
>     // Pass a variable from node to the browser
>     const url = isDev ? 'dev.example.com' : 'prod.example.com';
> 
>     await utils.runJS(
>       `
>       // Parameters get passed as an array into this context variable, and we can destructure them
>       const [url] = import.meta.pleasantestArgs
>       console.log(url)
>       // If we added a default exported function here, it would no longer be automatically called.
>       `,
>       // array of parameters passed here
>       [url],
>     );
>   }),
> );
> ```
> 
> This is a breaking change, because the previous mechanism for receiving parameters no longer works, and functions that are `default export`s from runJS are no longer called automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ test(
 
 #### Option 1: Rendering using a client-side framework
 
-If your app is client-side rendered, you can use [`utils.runJS`](#pleasantestutilsrunjscode-string-promisevoid) to tell Pleasantest how to render your app:
+If your app is client-side rendered, you can use [`utils.runJS`](#pleasantestutilsrunjscode-string-browserargs-unknown-promiserecordstring-unknown) to tell Pleasantest how to render your app:
 
 ```js
 import { withBrowser } from 'pleasantest';
@@ -367,7 +367,7 @@ Call Signatures:
 
 - `headless`: `boolean`, default `true`: Whether to open a headless (not visible) browser. If you use the `withBrowser.headed` chain, that will override the value of `headless`.
 - `device`: Device Object [described here](https://pptr.dev/#?product=Puppeteer&version=v13.5.2&show=api-pageemulateoptions).
-- `moduleServer`: Module Server options object (all properties are optional). They will be applied to files imported through [`utils.runJS`](#pleasantestutilsrunjscode-string-promisevoid) or [`utils.loadJS`](#pleasantestutilsloadjsjspath-string-promisevoid).
+- `moduleServer`: Module Server options object (all properties are optional). They will be applied to files imported through [`utils.runJS`](#pleasantestutilsrunjscode-string-browserargs-unknown-promiserecordstring-unknown) or [`utils.loadJS`](#pleasantestutilsloadjsjspath-string-promisevoid).
   - `plugins`: Array of Rollup, Vite, or WMR plugins to add.
   - `envVars`: Object with string keys and string values for environment variables to pass in as `import.meta.env.*` / `process.env.*`
   - `esbuild`: ([`TransformOptions`](https://esbuild.github.io/api/#transform-api) | `false`) Options to pass to esbuild. Set to false to disable esbuild.
@@ -1071,7 +1071,7 @@ Jest uses [jsdom](https://github.com/jsdom/jsdom) and exposes browser-like globa
   );
   ```
 
-- **No Synchronous DOM Access**: Because Jest runs your tests, Pleasantest will never support synchronously and directly modifying the DOM. While you can use [`utils.runJS`](#pleasantestutilsrunjscode-string-promisevoid) to execute snippets of code in the browser, all other browser manipulation must be through the provided asynchronous APIs. This is an advantage [jsdom](https://github.com/jsdom/jsdom)-based tests will always have over Pleasantest tests.
+- **No Synchronous DOM Access**: Because Jest runs your tests, Pleasantest will never support synchronously and directly modifying the DOM. While you can use [`utils.runJS`](#pleasantestutilsrunjscode-string-browserargs-unknown-promiserecordstring-unknown) to execute snippets of code in the browser, all other browser manipulation must be through the provided asynchronous APIs. This is an advantage [jsdom](https://github.com/jsdom/jsdom)-based tests will always have over Pleasantest tests.
 
 ### Temporary Limitations
 
--- a/README.md
+++ b/README.md
@@ -691,7 +691,7 @@ test(
 );
 ```
 
-The code string passed to `runJS` is also a module, and it can export values to make them available in Node. `runJS` returns a Promise resolving to the exports from the module that executed in the browser. Each export is wrapped in a `[`JSHandle`](https://pptr.dev/#?product=Puppeteer&version=v13.5.2&show=api-class-jshandle) (a pointer to an in-browser JS object), so that it can be passed back into the browser if necessary, or deserialized in Node using `.jsonValue()`.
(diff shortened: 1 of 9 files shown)


src/extend-expect.ts
```diff
@@ -197,9 +197,6 @@ Received ${this.utils.printReceived(arg)}`,
           () => matcher.call(this, elementHandle, ...matcherArgs),
           matchers[methodName],
         );
-        // AddHook resolves to undefined if the function throws after the async hook tracker closes
-        // Because it needs to not trigger an unhandled promise rejection
-        if (res === undefined) return { pass: !this.isNot, message: () => '' };
```
calebeby, line 202: The changes in `async-hooks.ts` explain why this situation won't happen.

spaceninja (approved):
These code changes look good and align with what we talked about in the dev team meeting.

+1 from me, but I'd appreciate it if you get a +1 from @gerardo-rodriguez as well, since some of this stuff is a little over my head, and he's more familiar with the scenario that led to this change.

README.md
```diff
@@ -678,17 +679,73 @@ test(
...
+      const [element] = import.meta.pleasantestArgs;
+      console.log(element);
       `,
       [element],
     );
   }),
 );
 ```
 
+The code string passed to `runJS` is also a module, and it can export values to make them available in Node. `runJS` returns a Promise resolving to the exports from the module that executed in the browser. Each export is wrapped in a `[`JSHandle`](https://pptr.dev/#?product=Puppeteer&version=v13.5.2&show=api-class-jshandle) (a pointer to an in-browser JS object), so that it can be passed back into the browser if necessary, or deserialized in Node using `.jsonValue()`.
```
gerardo-rodriguez, line 694:
```suggestion
The code string passed to `runJS` is also a module, and it can export values to make them available in Node. `runJS` returns a Promise resolving to the exports from the module that executed in the browser. Each export is wrapped in a [`JSHandle`](https://pptr.dev/#?product=Puppeteer&version=v13.5.2&show=api-class-jshandle) (a pointer to an in-browser JS object), so that it can be passed back into the browser if necessary, or deserialized in Node using `.jsonValue()`.
```

tests/utils/runJS.test.tsx
```diff
@@ -118,6 +127,44 @@ describe('Waiting for Promises in executed code', () => {
...
+        return ['it worked!', arg1, arg2]
+      }
+    `);
+    expect(Object.keys(result).sort()).toStrictEqual(['a', 'b', 'default']);
+    expect(result.a.toString()).toEqual('JSHandle:25');
+    expect(result.b.toString()).toEqual('JSHandle@function');
+    expect(result.default.toString()).toEqual('JSHandle:hi');
+
+    expect(await result.a.jsonValue()).toStrictEqual(25);
+    expect(await result.b.jsonValue()).toStrictEqual({}); // function is not JSON-serializable
```
gerardo-rodriguez, line 153: Thanks for the comment! 👍🏽

gerardo-rodriguez (approved):
This is great work, @calebeby, yay! 🎉 

And the documentation is _excellent!_ Thank you so much. 💯

PR #560, opened 2022-08-29
Update puppeteer and fix types

calebeby:
- I updated to puppeteer 16 but the latest is 17 (came out today). We can't update to it yet since `@axe-core/puppeteer` hasn't updated their peerDep yet.
- Puppeteer has a new docs site. They changed their URL structure. I fixed all the links in the README.
- Puppeteer changed the way they do types for `evaluate` and `evaluateHandle`. I updated our types to work with that (no user-facing changes except if the user is using `evaluate`/`evaluateHandle` directly).

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -16,13 +16,6 @@
       "rangeStrategy": "widen"
     },
     {
-      "matchPackageNames": ["puppeteer"],
-      "prBodyNotes": [
-        ":warning: **Warning:** Manually update the version numbers in URL's in README"
-      ],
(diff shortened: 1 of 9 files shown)


spaceninja:
For testing, is it sufficient to see that the checks on this PR passed?

spaceninja (approved):
Code changes look good and checks pass, so I'm approving, but if you need a more thorough check, holler at me.

calebeby:
> For testing, is it sufficient to see that the checks on this PR passed?

Yep! Pleasantest has good tests ✅

PR #640, opened 2023-01-30
Update puppeteer

calebeby:
I needed to update puppeteer to fix an issue with updating pleasantest on c4/patterns. The latest puppeteer is v19, but @axe-core/puppeteer doesn't "support" it yet. v18 includes the fix we need, so I am going to release a version with that.


README.md
```diff
@@ -669,7 +669,7 @@ test(
 );
 ```
 
-To pass variables from the test environment into the browser, you can pass them in an array as the 2nd parameter. Note that they must either be JSON-serializable or they can be a [`JSHandle`](https://pptr.dev/api/puppeteer.jshandle) or an [`ElementHandle`](https://pptr.dev/#?product=Puppeteer&version=v13.0.0&show=api-class-elementhandle). The arguments will be received in the browser via `import.meta.pleasantestArgs`:
+To pass variables from the test environment into the browser, you can pass them in an array as the 2nd parameter. Note that they must either be JSON-serializable or they can be a [`JSHandle`](https://pptr.dev/api/puppeteer.jshandle) or an [`ElementHandle`](https://pptr.dev/api/puppeteer.elementhandle/). The arguments will be received in the browser via `import.meta.pleasantestArgs`:
```
calebeby, line 672: This is unrelated but it was a link I missed changing when I went through to switch the links to their new docs site

package.json
```diff
@@ -9,20 +9,16 @@
   ],
   "license": "MIT",
   "peerDependencies": {
-    "@axe-core/puppeteer": "^4.4.4",
-    "axe-core": "^4.4.3"
+    "@axe-core/puppeteer": "^4.4.4"
```
calebeby, line 12: I kept this version range wider to avoid a breaking change (but it still kinda sorta is since the newer version of @axe-core/puppeteer is needed to make npm allow the newer puppeteer version, although it probably would work fine with passing `--force` to npm)

package.json
```diff
@@ -9,20 +9,16 @@
...
       "optional": true
-    },
-    "axe-core": {
-      "optional": true
     }
   },
   "devDependencies": {
     "@ampproject/remapping": "2.2.0",
-    "@axe-core/puppeteer": "4.4.5",
+    "@axe-core/puppeteer": "4.6.0",
```
calebeby, line 21: Needed to satisfy the peerDep since puppeteer was updated

src/accessibility/index.ts
```diff
@@ -50,7 +50,7 @@ const getAccessibilityTree = async (
     // eslint-disable-next-line @cloudfour/typescript-eslint/no-unnecessary-condition
     element &&
     typeof element === 'object' &&
-    element.constructor.name === 'Page'
+    element.constructor.name === 'CDPPage'
```
calebeby, line 53: Newer version of Puppeteer changes the name of this class

src/accessibility/index.ts
```diff
@@ -239,7 +239,7 @@ async function toPassAxeTests(
   } catch {
     throw removeFuncFromStackTrace(
       new Error(
-        'Install @axe-core/puppeteer and axe-core to use the toPassAxeTests matcher',
+        'Install @axe-core/puppeteer to use the toPassAxeTests matcher',
```
calebeby, line 242: I realized that @axe-core/puppeteer comes with its own version of axe-core so it is not necessary for users to directly install axe-core.

tests/accessibility/toPassAxeTests.test.ts
```diff
@@ -13,7 +13,7 @@ test(
       Violations found:
 
       [31m[1mImages must have alternate text[22m[39m (image-alt)
-      https://dequeuniversity.com/rules/axe/4.4/image-alt?application=axe-puppeteer
+      https://dequeuniversity.com/rules/axe/4.6/image-alt?application=axe-puppeteer
```
calebeby, line 16: From updating @axe-core/puppeteer, axe-core was updated, so their URLs updated too.

spaceninja (approved):
Please make sure your commit message reflects your comments here about why we no longer need to install axe-core directly.

PR #726, opened 2023-08-22
Bugfix: fix URL path for childProcess.fork when using Pleasantest from ESM

calebeby:
`childProcess.fork` does not accept a `file://` path. It requires a regular path. The `resolveFileUrl` function I put in the rollup config returned a `file://` path for ESM bundles and a regular path for CJS bundles. This meant that CJS bundles worked but ESM bundles didn't. This PR removes the `resolveFileUrl` in favor of rollup's built-in version, which always returns a `file://` path. Then when we pass it to `childProcess.fork` we use `fileURLToPath` to switch it back.

Unfortunately we don't have any test cases for using Jest's experimental ESM with pleasantest. I manually created a repo to test this. I hope to add a monorepo setup to pleasantest so that we can create test cases with Jest's experimental ESM without forcing us to switch all our tests over.

### DIFF
--- a/.changeset/calm-windows-retire.md
+++ b/.changeset/calm-windows-retire.md
@@ -1,0 +1,5 @@
+---
+'pleasantest': minor
+---
+
+Fix fork URL for ESM use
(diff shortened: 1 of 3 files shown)


Paul-Hebert (approved):
These changes seem reasonable to me, especially if you've tested them in another repo :+1:

PR #732, opened 2023-08-23
Allow returning non-serializable values from `waitFor`

calebeby:


tests/wait-for.test.ts
```diff
@@ -40,7 +63,9 @@ test(
       tests/wait-for.test.ts
 
               throw new Error('something bad happened');
-                    ^"
+                    ^
+      -------------------------------------------------------
+      dist/cjs/index.cjs"
```
Paul-Hebert, line 68: Just double checking that it was expected for this test's output to change as well.

Paul-Hebert (approved):
I don't have enough context to 100% understand these changes but I don't see any obvious issues and they seem reasonable to me.

tests/wait-for.test.ts
calebeby: I am not entirely sure why this changed (not sure why it was how it was before this PR) but yes this is the expected output

PR #748, opened 2024-07-22
Make it work on Windows

calebeby:
Hello and welcome to my pull request. The following are things about Windows that make me sad.

- ☹️ Windows paths start with `C:\` (or similar). On Windows, we need to handle those as absolute paths.
- ☹️ Windows paths (usually) have `\` instead of `/`. On Windows, we need to use `\` for any path used on the file system, and `/` for any path used in the browser. In a lot of cases Windows doesn't seem to care which one you use, but I don't know if that is universally true (and if you _ask_ for a path it will always give you one with `\` AFAICT)
- ☹️ Windows files have `\r\n` (CRLF) instead of `\n` between each line. This is not always true though; for example git for windows can be configured to checkout files as-is from GitHub instead of converting them while performing a checkout. So they can have either `\r\n` or `\n`. But text files in the browser (e.g. HTML, CSS, JS) will have `\n` instead since they are not on the filesystem. So we need to handle both.
- ☹️ The fact that windows paths usually use `\` which happens to be the escape character for JS strings makes things more difficult, e.g. you can't plop a windows path into a JS string with `${}` directly because of the escaping. Don't ask how long I spent debugging a string that once interpolated became `something\node_modules` and the `\n` in the middle was seen as a line break so then I was like "what is ode modules I do not like it" 😱 🙀 😡 :rage4: 
- ☹️ Windows paths are case insensitive so they must be normalized before they can be compared


package.json
```diff
@@ -118,13 +118,15 @@
...
-      "types": "./dist/index.d.mts"
+    ".": {
+      "require": {
+        "default": "./dist/cjs/index.cjs",
+        "types": "./dist/index.d.cts"
+      },
+      "import": {
+        "default": "./dist/esm/index.mjs",
+        "types": "./dist/index.d.mts"
+      }
```
calebeby, line 129: I don't know why this change was necessary, but windows is happy with this, whereas it was not happy with the previous code.

src/module-server/bundle-npm-module.ts
```diff
@@ -70,10 +70,9 @@ export const bundleNpmModule = async (
           },
           load(id) {
             if (id === virtualEntry) {
-              const code = `export * from '${mod}'
-export {${namedExports.join(', ')}} from '${mod}'
-export { default } from '${mod}'`;
-              return code;
+              return `export * from ${JSON.stringify(mod)}
+              export {${namedExports.join(', ')}} from ${JSON.stringify(mod)}
+              export { default } from ${JSON.stringify(mod)}`;
```
Paul-Hebert, line 75: Does it make sense to stringify it once?

```suggestion
              const normalizedMod = JSON.stringify(mod);
              return `export * from ${normalizedMod}
              export {${namedExports.join(', ')}} from ${normalizedMod}
              export { default } from ${normalizedMod}`;
```

(Not sure about var name.)

src/module-server/middleware/js.ts
```diff
@@ -102,7 +102,14 @@ export const jsMiddleware = async ({
...
         const fileSrc = await fs.readFile(file, 'utf8');
-        const inlineStartIdx = fileSrc.indexOf(code);
+        let EOL = '\n';
+        // Find the first line end (\n) and check if the character before is \r
+        // This tells us whether the file uses \r\n or just \n for line endings.
+        // Using node:os.EOL is not sufficient because git on windows
+        // Can be configured to check out files with either kind of line ending.
+        const firstLineEndPos = fileSrc.indexOf('\n');
+        if (fileSrc[firstLineEndPos - 1] === '\r') EOL = '\r\n';
+        const inlineStartIdx = fileSrc.indexOf(code.replaceAll('\n', EOL));
```
Paul-Hebert, line 112: I wonder if some of this should be pulled out into a helper called `normalizeLineEndings` or `determineLineEndings` or something?

src/module-server/node-resolve.ts
```diff
@@ -129,7 +129,7 @@ export const resolveFromNodeModules = async (
   const cacheKey = resolveCacheKey(id, importer, root);
   const cached = resolveCache.get(cacheKey);
   if (cached) return cached;
-  const pathChunks = id.split(posix.sep);
+  const pathChunks = id.split(/[/\\]/g);
```
Paul-Hebert, line 132: Can we add a comment explaining what this regex is looking for and how it works?

tests/test-utils.ts
```diff
@@ -48,12 +50,19 @@ const stripAnsi = (input: string) => input.replace(ansiRegex(), '');
 
 const removeLineNumbers = (input: string) => {
   const lineRegex = /^\s*▶?\s*(\d)*\s+│/gm;
-  const fileRegex = new RegExp(`${process.cwd()}([a-zA-Z/._-]*)[\\d:]*`, 'g');
+  const fileRegex = new RegExp(
```
Paul-Hebert, line 53: Can we explain this regex as well or provide a link to more info?

Paul-Hebert (approved):
Thanks Caleb! Code looks good and tests are passing :+1:

I left a few small non-blocking suggestions that you may want to take a look at before merging. Thanks!

